PR #39, opened 2022-07-17
Test panel: add support for accounts

### LINKED ISSUE #36: [Feature Request] UI to fetch account data
I think the test UI would be even more useful if there was a way to fetch the account data from the deployed program and view it in the playground.

I think a way to do `program.account.newAccount.fetch(publicKey)` and `program.account.newAccount.all()` would add a huge amount of value. This would let us test an instruction and immediately see its effect on-chain. 

I expect the UI for this would allow selecting an account (parsed from the IDL), either selecting all or optionally entering a specific address, and it'd then display the fetched data as nicely formatted JSON.

mcintyre94:
This PR updates the Test panel to add support for displaying the accounts parsed from the IDL, and for each of them fetching all accounts or one account by its public key.

This allows inspecting the results of an instruction immediately and without leaving the Test UI. 

Demo: 
https://user-images.githubusercontent.com/1711350/179417905-1f1794cf-a95b-4e0c-880e-5ffb1a201c2c.mov

Implementation details:

- I've tried to keep the styling consistent with the existing instructions. I'm not experienced with styled components though so please let me know if I did anything wrong! 
- The new component used to display a row of `Account` data (in a Foldable, similar to instructions) is `FetchableAccount`
- There's a bit of a hack to override the `toJson` method on the `BN` object. By default it's a hex string (https://github.com/indutny/bn.js/#utilities) which isn't very helpful when it's a public key. I tried implementing a replacer in `JSON.stringify` but that didn't work, the default toJson has already been called on it before it's passed to the replacer. So I override the prototype to use `toString` instead, which gives us the public keys we expect.
- Similarly to with instructions, the actual account fetching is implemented in static functions on a new `PgAccount` class.
- The account names need to be camel-case when used in `program.account.newAccount.[...]`, whereas they're in pascal case in the IDL.

Screenshot 2022-07-17 at 18 25 35

Closes #36

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,6 @@
 server
 
 # wasm
-*/target
+**/target
 **/www
 
--- a/client/src/components/Panels/Side/Right/Test/FetchableAccount.tsx
+++ b/client/src/components/Panels/Side/Right/Test/FetchableAccount.tsx
@@ -1,0 +1,129 @@
+import { BN, Idl } from "@project-serum/anchor";
+import { useConnection } from "@solana/wallet-adapter-react";
+import { PublicKey } from "@solana/web3.js";
+import { FC, useEffect, useState } from "react";
+import styled, { css } from "styled-components";
+import { PgAccount } from "../../../../../utils/pg/account";
+import Button from "../../../../Button";
+import Foldable from "../../../../Foldable";
+import Input, { defaultInputProps } from "../../../../Input";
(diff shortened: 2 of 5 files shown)


acheroncrypto (approved):
This is awesome, thanks for your effort!

I think it would be better if we handled the error cases, please let me know if you want to work on that too. Totally fine if you don't want to.

PR #51, opened 2022-08-16
Add support for compiling Python files using Seahorse

mcintyre94:
Draft because:

- ~~Needs the program ID fix from #50 (or equivalent)~~
- ~~Currently uses a fork of seahorse. We probably want to wait for that to merge first? PR to track: https://github.com/ameliatastic/seahorse-lang/pull/10~~ Updated!!

The code here should be in a good state to review though :) 

This PR adds a new wasm package `seahorse-compile`. This is a very thin wrapper around the `seahorse-lang` crate, which just gives us a wrapper around the `compile` function that returns a string for wasm. It also adds error handling into the Playground terminal using the playground-utils:
Screenshot 2022-08-16 at 20 56 17

Initial support for Python is also added to the client:

- If there are `*.py` files, they are compiled with the seahorse wasm function and then passed to the build API.
- The logic that updates program IDs pre-build, in the editor, and in localstorage are all modified to support Python files in addition to Rust

The result is that we have the ability to build, deploy and test a Seahorse Python file exactly the same as a Rust one 🎉 

https://user-images.githubusercontent.com/1711350/184974747-4e070020-9f65-4d8f-b92d-e88c8303886f.mp4

This PR is quite minimal, and doesn't include any of the following:

- There's no indication in the UI that Seahorse/Python is available. You need to add/rename a `.py` file yourself. Not sure exactly how/where we'd want to message this 
- There's no syntax highlighting for Python yet. I did get this working pretty easily in my earlier fork, but I replaced Rust rather than making it switch between/support both which we'd need to do here. Probably not too hard though! Relevant commit: https://github.com/mcintyre94/solana-playground/commit/c4378320120ee700a7fab52167fc737c1ed32cb7 
- There's no Python icon by `.py` files yet
- For user friendliness, we might want to figure out a way to filter out the rust warnings on compile. Seahorse generates a lot of them and users who don't know Rust might find them a bit scary!


client/src/components/Panels/Main/Editor/Editor.tsx
```diff
@@ -252,51 +252,108 @@ const Editor = () => {
     //eslint-disable-next-line react-hooks/exhaustive-deps
   }, [editor, explorer, explorerChanged, setNoOpenTabs]);
 
+  const updateLibRs = (currentContent: string, programId: string) => {
```
acheroncrypto, line 255: `updateLibRs` and `updateLibPy` seems to be nearly identical functions. Would be nice to have a single function with an identifier argument.

client/src/utils/pg/build.ts
```diff
@@ -13,17 +14,51 @@ interface BuildResp {
...
   static async build(files: Files) {
     const programInfo = PgProgramInfo.getProgramInfo();
 
+    const pythonFiles = files.filter(([fileName]) =>
+      fileName.toLowerCase().endsWith(".py")
+    );
+
+    let rustFiles: Files;
+    if (pythonFiles.length > 0) {
+      const seahorsePkg = await PgPkg.loadPkg(PkgName.SEAHORSE_COMPILE);
```
acheroncrypto, line 32: This line will re-import the seahorse wasm crate every time it runs. Would be nice to have some sort of in-memory access to this function after the first import similar to https://github.com/solana-playground/solana-playground/blob/master/client/src/components/Panels/Main/Terminal/usePkg.tsx#L27

client/src/utils/pg/build.ts
```diff
@@ -13,17 +14,51 @@ interface BuildResp {
...
     const resp = await fetch(`${SERVER_URL}/build`, {
       method: "POST",
       headers: {
         "Content-Type": "application/json",
       },
       body: JSON.stringify({
-        files,
+        files: rustFiles,
         uuid: programInfo.uuid,
+        kp: programInfo.kp,
```
acheroncrypto, line 61: No need for this anymore

client/src/components/Panels/Main/Editor/Editor.tsx
mcintyre94: Good shout, I've refactored this to a single function. Definitely should be easier to maintain now!

client/src/utils/pg/build.ts
mcintyre94: TY - dropped that commit

client/src/utils/pg/build.ts
mcintyre94: Nice, I've refactored this a bit. The package state is now maintained in `useBuild`, and the logic to determine if we have Python files is there too. So the package will only be loaded the first time a build with Python files happens.

I've also refactored in this file to a `buildPython` and `buildRust` function, where the `buildPython` ones takes the package as input, compiles to rust and then calls/returns `buildRust`. I think that's a bit cleaner too, it removes any package loading from here and better shows how the Python and Rust builds relate to each other.

acheroncrypto: Yes, it's much better now!

acheroncrypto (approved):

PR #57, opened 2022-09-13
Fix nightly builds

BorovkovSergey:
solana-extra crate doesn't build with nightly toolchain. 

* The main problem was in [solana-frozen-abi](https://github.com/solana-labs/solana/blob/master/frozen-abi/build.rs) It was compiling without needed feature - `specialization`.
* also was added log crate that needed for some macroses inside `solana-frozen-abi`


acheroncrypto (approved):
Looks good, thank you!

PR #62, opened 2022-09-28
Add keypair to the wallet interface so it can be used in client code

mcintyre94:
This PR slightly modifies the `PgWallet` interface to make `keypair` accessible.

This allows writing client code like:
```ts
const transactionSignature = await web3.sendAndConfirmTransaction(
  pg.connection,
  transaction,
  [pg.wallet.keypair]
);
```
This is a lot simpler than using `pg.wallet.signTransaction` and then sending it as a raw transaction. I think that's currently how you'd need to send a transaction?

Note that this doesn't have any security implications, client code can already run `pg.wallet._kp` to get at the currently private field.

If we do need to hide the keypair for security then maybe we could add `_kp` to the blacklisted words? But I do think it's a lot nicer to be able to access the keypair in client code if possible!

### DIFF
--- a/client/src/components/Panels/Main/Editor/Monaco/declarations/pg.raw.d.ts
+++ b/client/src/components/Panels/Main/Editor/Monaco/declarations/pg.raw.d.ts
@@ -2,6 +2,7 @@ declare module "solana-playground" {
   import * as web3 from "@solana/web3.js";
   interface PgWallet {
     publicKey: web3.PublicKey;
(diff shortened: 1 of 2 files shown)


acheroncrypto (approved):
I originally wanted to only allow browser way of doing things since we are running the code on the browser but now I think it's much better to allow both ways so the example NodeJS codes run without needing any changes.

PR #65, opened 2022-10-14
Update seahorse to 0.2.1 and handle multiple output rust files

mcintyre94:
As of 0.2.0 Seahorse is rewritten and now outputs multiple Rust files, instead of just the single rust source file that `compile` used to return. This PR adds support for this updated Seahorse output.

Specifically, `compile` now returns a `Tree` data structure, where the nodes are the path to a rust source file, starting from inside the `src/` directory. The leaves are source code for a single file each. For example there can be `dot` -> `mod` -> (source code for `dot/mod.rs`)

Unfortunately WASM doesn't let us return this sort of structured data. The best option I found was to return a `Vec<JsValue>`, where each `JsValue` is a string. I tried doing nested `Vec<Vec<JsValue>>` and making `Vec<String>` the output type, it didn't like either.

So we return an array like this: `["dot/mod.rs", "(source code of dot/mod.rs)", "lib.rs", "(source code of lib.rs")]`

There's new code in `wasm/seahorse-compile/src/lib.rs` to perform this conversion. We also prepend `src/` to all the paths because that's what the Playground build API expects.

On the Javascript side, we already basically want this `[path, content]` format, we just need to make it a 2d array. We convert `[path1, content1, path2, content2]` to `[[path1, content1], [path2, content2]]` before sending it to `buildRust`. We also still need to replace `# [program]` with `#[program]` since Seahorse called from WASM doesn't format the rust output.


acheroncrypto (approved):
Looks great, thank you so much!

PR #66, opened 2022-10-17
Use the same import regex when removing imports

mcintyre94:
This fixes a bug where code after the import was also being removed


acheroncrypto (approved):
Makes sense, thank you!

PR #78, opened 2022-11-11
Add events to the Test page, display recorded events from the program

### LINKED ISSUE #73: [Feature] Add Anchor event listeners to test UI
Something that I think would be helpful in the test UI is the ability to see Anchor events that happen as a result of the instructions that you perform. You could do this with client code (add listeners that log events, run instructions), but I think it'd be nice as part of the test UI.

I think a potential implementation would be:

- Add a listener for each event in the IDL on the `test` page. We'd only record events while the test page is empty, I think that's fine.

- Add a new subsection titled 'Events', with a foldable section (like we have for each instruction/account) for each event. When expanded it would show any instances of that event that have been recorded while the test page was open as formatted JSON. Probably a 'clear' button would be nice too, which just removes the events of that type so you can start again. 

This would allow writing code that uses Anchor events, and then easily debugging the events emitted by various calls to that instruction.

mcintyre94:
This PR adds an events section to the test page. It lists the events in the anchor IDL, listens for instances of those events and displays them in a foldable section.

https://user-images.githubusercontent.com/1711350/201411651-efe85006-fa13-48b7-b63a-1e648d1ce605.mov

This means that you can verify/experiment with the events emitted by your program while testing it in playground.

I've refactored out some stuff that's shared with the accounts section: the styled code block, and the hacky replacement of `BN.toJson` to make large numbers display nicely.

Note that the event listeners are only active when the test page is open, they're cleaned up on navigation away and the recorded events are reset too. 

Closes #73

### DIFF
--- a/client/src/components/Panels/Side/Right/Test/Test.tsx
+++ b/client/src/components/Panels/Side/Right/Test/Test.tsx
@@ -7,9 +7,14 @@ import FetchableAccount from "./FetchableAccount";
 import Text from "../../../../Text";
 import TestSkeleton from "./TestSkeleton";
 import { ConnectionErrorText } from "../Common";
-import { PgProgramInfo } from "../../../../../utils/pg";
+import { PgProgramInfo, PgTest } from "../../../../../utils/pg";
 import { buildCountAtom } from "../../../../../state";
 import { useInitialLoading } from "../";
+import { useMemo } from "react";
+import { useConnection } from "@solana/wallet-adapter-react";
+import { useCurrentWallet } from "../../../Wallet";
+import ListenableEvent from "./ListenableEvent";
+import { useBigNumberJson } from "../../../../../hooks/useBigNumberJson";
 
 // Webpack 5 doesn't polyfill buffer
 window.Buffer = buffer.Buffer;
@@ -21,6 +26,17 @@ const Test = () => {
   const { initialLoading, deployed, connError } = useInitialLoading();
 
   const idl = PgProgramInfo.getProgramInfo()?.idl;
+  const { connection: conn } = useConnection();
+  const { currentWallet } = useCurrentWallet();
+
+  const program = useMemo(() => {
+    return idl && currentWallet
+      ? PgTest.getProgram(idl, conn, currentWallet)
+      : null;
+  }, [idl, conn, currentWallet]);
+
+  // Used for both accounts and events data
+  useBigNumberJson();
 
   if (initialLoading) return <TestSkeleton />;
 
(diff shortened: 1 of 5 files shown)


client/src/components/Panels/Side/Right/Test/Test.tsx
```diff
@@ -20,7 +24,50 @@ const Test = () => {
 
   const { initialLoading, deployed, connError } = useInitialLoading();
 
-  const idl = PgProgramInfo.getProgramInfo()?.idl;
+  const idl = useMemo(() => PgProgramInfo.getProgramInfo()?.idl, []);
```
acheroncrypto, line 27: This change will result in IDL not getting updated if you `build` from the terminal when the test section is visible. You can either remove the `useMemo` or use `buildCount` as a depeendency from
```ts
const [buildCount] = useAtom(buildCountAtom)
```

I think it's probably better to not use `useMemo` at all because the parent component is not getting rendered a lot anyway but I'm open to both changes.

client/src/components/Panels/Side/Right/Test/Test.tsx
```diff
@@ -20,7 +24,50 @@ const Test = () => {
...
+    return idl && currentWallet
+      ? PgTest.getProgram(idl, conn, currentWallet)
+      : null;
+  }, [idl, conn, currentWallet]);
+
+  const [receivedEvents, setReceivedEvents] = useState<{
+    [eventName: string]: Object[];
+  }>({});
+
+  const handleEvent = (eventName: string, event: Object) => {
```
acheroncrypto, line 41: Is there a reason why `Event` related code needs to live in the `Test` component instead of its own component as a child component of `Test` like the other program interactions?

client/src/hooks/useBigNumberJson.tsx
```diff
@@ -0,0 +1,18 @@
+import BN from "bn.js";
+import { useEffect } from "react";
+
+export const useBigNumberJson = () => {
```
acheroncrypto, line 4: Maybe this can only be used in the parent component instead of every child component?

acheroncrypto (changes requested):

client/src/components/Panels/Side/Right/Test/Test.tsx
mcintyre94: Thanks, I've removed the `useMemo` and checked that it updates now when you run `build` in the terminal.

client/src/components/Panels/Side/Right/Test/Test.tsx
mcintyre94: Nope, I think I did it here originally and then when I refactored thought the event listeners needed to stay here. But since there's a separate listener per event I think it's better to just put them in `ListenableEvent`. Also simplifies things because it's all in the one component so no need to map by event name any more.

client/src/hooks/useBigNumberJson.tsx
mcintyre94: Agreed, I've moved it to just Test

acheroncrypto (approved):

PR #89, opened 2022-11-29
Vscode: Fix seahorse project generate

mcintyre94:
The seahorse project create was erroring on the check for an existing `programs/` directory, because it initializes an Anchor project first. The nesting when there are existing files also isn't appropriate for seahorse.

This PR updates the Seahorse create to write Seahorse files directly after the Anchor create. This avoids the checks and nesting of `processCreate` which aren't suitable when we know there are existing files (the Anchor ones).

### DIFF
--- a/vscode/src/commands/create/seahorse.ts
+++ b/vscode/src/commands/create/seahorse.ts
@@ -1,48 +1,46 @@
 import * as path from "path";
 
 import { PATHS } from "../../constants";
+import { Files, PgFs } from "../../utils";
 import { processCreateAnchor } from "./anchor";
-import { processCreate } from "./common";
 
 export const processCreateSeahorse = async () => {
   // Create Anchor
-  const { name } = await processCreateAnchor();
+  const { name, projectUri } = await processCreateAnchor();
 
-  // Create Seahorse
-  await processCreate({
-    name,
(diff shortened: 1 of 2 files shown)


vscode/src/commands/create/seahorse.ts
```diff
@@ -1,48 +1,48 @@
...
+      path.join(
+        PATHS.DIRS.PROGRAMS_PY,
+        PATHS.DIRS.SEAHORSE,
+        PATHS.FILES.PRELUDE_PY
+      ),
+      require("./seahorse-prelude.py"),
+    ],
+  ];
+
+  const baseUri = await PgFs.getBaseUri();
```
acheroncrypto, line 45: We should use the same `baseUri` that `processCreateAnchor` uses. Anchor could create a nested folder so there is no guarantee `baseUri` we get here is the same as the Anchor project uri.

acheroncrypto (changes requested):

vscode/src/commands/create/seahorse.ts
mcintyre94: Thanks! Updated to return a `projectUri` from `processCreateAnchor` which is the URI seahorse should use.

Issue #86, opened 2022-11-27
please update Seahorse version to 0.2.4


saikishore222:

acheroncrypto:
Updated on 0ef8e9e.

PR #90, opened 2022-11-30
Update seahorse to latest commit, which includes a wasm fix

### LINKED ISSUE #75: need to update the seahorse version for writing Smart Contracts in Seahorse

mcintyre94:
- Update Seahorse to the latest current commit. This is pre-release, but fixes the issue that stopped it compiling in wasm
- Slight change to lib.rs because the `compile` signature has changed a bit. We pass a `PathBuf` in, it's optional but defaults to finding the current directory - which is a runtime error in wasm
- I've bumped the toolchain version. I was getting an error that `2022-06-27` doesn't exist, and [it doesn't seem to be in releases](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1620-2022-06-30). Not sure what's going on there, but I've bumped to the latest version

Note that this still doesn't give us the multi-file python imports feature that seahorse now has. We're still using `compile` which will only work with a single Python source file. But it brings us up to date with the latest seahorse features/fixes within that existing constraint.

Fixes #75, #86


acheroncrypto:
Nice! I think it's good that we use the latest stable version anyway but have you tried to compile other wasm packages in the repo? We currently don't have Github actions checks and I'm afraid it could break other builds.

I'll merge if you can confirm other packages compile with these versions.

mcintyre94:
I haven't built most of these before so might have done something wrong, but I've tried to build them all with this change.
I also used the script from [this awesome stackoverflow answer](https://stackoverflow.com/a/10847242/1375972) to check for file size differences


### rustfmt/rustfmt-wasm
I got an error building this, but it seems to have its own toolchain and the error refers to that version. I haven't built this myself before so not sure if there's something I'm doing wrong

```sh
$ cd rustfmt/rustfmt-wasm/
$ wasm-pack build
Error: Error during execution of `cargo metadata`: info: syncing channel updates for 'nightly-2018-09-07-aarch64-apple-darwin'
info: latest update on 2018-09-07, rust version 1.30.0-nightly (a8c11d216 2018-09-06)
error: target 'aarch64-apple-darwin' not found in channel.  Perhaps check https://doc.rust-lang.org/nightly/rustc/platform-support.html for available targets
```


### solana-cli
no error, output is a bit smaller
```sh
$ ../../diff-sizes.sh --cached HEAD | grep solana-cli
-1      wasm/pkgs/solana-cli/package.json
2168    wasm/pkgs/solana-cli/solana_cli_wasm_bg.js
-93642  wasm/pkgs/solana-cli/solana_cli_wasm_bg.wasm
0       wasm/pkgs/solana-cli/solana_cli_wasm_bg.wasm.d.ts
```


### solana-client
I got an error saying this isn't the right crate-type for wasm
```sh
$ wasm-pack build
Error: crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:

[lib]
crate-type = ["cdylib", "rlib"]
```

After adding that it built with no error
Doesn't seem to be a built version currently checked in to compare against


### spl-token-cli
no error, a bit smaller too
```sh
$ ../../diff-sizes.sh --cached HEAD | grep spl-token-cli
2177    wasm/pkgs/spl-token-cli/spl_token_cli_wasm_bg.js
-85504  wasm/pkgs/spl-token-cli/spl_token_cli_wasm_bg.wasm
0       wasm/pkgs/spl-token-cli/spl_token_cli_wasm_bg.wasm.d.ts
```


### sugar-cli
no error, slightly bigger
```sh
$ ../../diff-sizes.sh --cached HEAD | grep sugar-cli    
0       wasm/pkgs/sugar-cli/sugar_cli_wasm_bg.js
16653   wasm/pkgs/sugar-cli/sugar_cli_wasm_bg.wasm
0       wasm/pkgs/sugar-cli/sugar_cli_wasm_bg.wasm.d.ts
```

PR #98, opened 2022-12-07
Update Seahorse to 0.2.5 + add an update script for wasm dependencies

### LINKED ISSUE #95: Update Seahorse to 0.2.5
Seahorse 0.2.5 was just released

We're already pretty close to 0.2.5 because we needed a very recent fix, so I expect this will be an easy update

Feel free to assign to me :)

mcintyre94:
Automating stuff is better than doing stuff, so I added a new script `wasm/update.sh` that does the update for us

It does the `wasm-pack build` and then copies all generated files (except gitignore) into the `pkgs/<package>` directory, and then updates the `yarn` local dependency pointing at it. Should work for any package that uses that workflow to update, which I think is all of them that are dependencies of `client`.

Closes #95


acheroncrypto (approved):
This script is awesome and much appreciated, thank you!

PR #113, opened 2023-02-04
Fix buggy seed inputs and add additional tutorial docs

fletchertyler914:
Screenshot 2023-02-04 at 12 09 02 AM

Currently there is no good way to generate an account from seed without the first seed being a string, as seen above.
Leaving the `Seed(1): string` empty and adding a new `Seed(2): Publickey` lets me generate the correct account, but it feels buggy and unintuitive. 

I propose a solution by only showing the remove seed button when there are more than one seed, and moving the add seed button down below the seeds.

Screenshot 2023-02-04 at 12 16 20 AM

This allows you to adda new signer more intuitively, and now gives you an option of removing any seed as long as there is at least one. The default state still does not allow removing the default signer:

Screenshot 2023-02-04 at 12 17 32 AM

Because of these UI changes, I went ahead and finished the tutorial walkthrough and provided step by step screenshots of the new UI to make it easy to follow along. You can see those updates in the tutorial pages.

Let me know if there's anything missing! 

Also, thanks so much for this repo. Its a wonderful tool for the Solana community. 
I will be writing about it on [Substack](https://lostsols.substack.com/?showWelcome=true) in the future.


fletchertyler914:
Just updated this to include information about the client test in the explorer window again. That provides walkthroughs on both testing options (test files or interactive test window).

acheroncrypto:
Hey, I agree the this case is not intiutive especially for beginners. I was planning to add a feature to switch the existing seed types by clicking the type(`string` in our case) but this solution works too.

Thank you so much, I appreciate it!

acheroncrypto (commented):
I like the idea of showcasing the steps but it also has some cons. It could get outdated pretty quickly and the amount of screenshots is quite a lot. What do you think of reducing some of the screenshots or making a short <10s video instead of many screenshots to showcase the test UI for example?

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #119, opened 2023-03-13
Add `simulate_transaction` methods to wasm solana client 

Eliascm17:
- added both `simulate_transaction_with_config` and `simulate_transaction` methods to wasm solana client
- updated `SimulateTransactionResponse` to return the [object](https://docs.solana.com/api/http#simulatetransaction) that is expected from the `simulateTransaction` Http method instead of `Signature` type. Mirroring what is found in the [Solana RPC Client](https://github.com/solana-labs/solana/blob/6e5615e32dfd20730f9c3a65bb211512d867e103/rpc-client-api/src/response.rs#L416-#L424)

### DIFF
--- a/wasm/Cargo.lock
+++ b/wasm/Cargo.lock
@@ -2861,10 +2861,9 @@ dependencies = [
 
 [[package]]
 name = "solana-playground-utils-wasm"
-version = "0.1.0"
+version = "1.11.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


acheroncrypto (approved):
Not sure if this works without `#[serde(rename_all = "camelCase")]` but I've added them just in case.

PR #121, opened 2023-03-24
Upgrade Solana to 1.14.16

nickgarfield:
- Upgrades Solana dependencies to 1.14.16.
- Upgrades `anchor-cli-wasm` version to 0.27.0 to mirror the latest Anchor release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 | [num-traits](https://docs.rs/num-traits/0.2.15)                                    | 0.2.15  |
 | [pyth-sdk](https://docs.rs/pyth-sdk/0.7.0)                                         | 0.7.0   |
 | [pyth-sdk-solana](https://docs.rs/pyth-sdk-solana/0.7.0)                           | 0.7.0   |
-| [solana-program](https://docs.rs/solana-program/1.14.16)                           | 1.14.16 |
+| [solana-program](https://docs.rs/solana-program/1.14.12)                           | 1.14.12 |
 | [spl-associated-token-account](https://docs.rs/spl-associated-token-account/1.1.2) | 1.1.2   |
 | [spl-token](https://docs.rs/spl-token/3.5.0)                                       | 3.5.0   |
 | [switchboard-v2](https://docs.rs/switchboard-v2/0.1.20)                            | 0.1.20  |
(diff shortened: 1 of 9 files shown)


wasm/solana-client/Cargo.toml
```diff
@@ -22,8 +22,8 @@ reqwest = { version = "0.11.9", features = ["json"] }
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
-solana-extra-wasm = { path = "../utils/solana-extra", version = "=1.14.15" }
-solana-sdk = "=1.14.15"
+solana-extra-wasm = { path = "../utils/solana-extra", version = "1.14.16" }
+solana-sdk = "1.14.16"
```
acheroncrypto, line 26: Is there a reason why `=` was removed from the versions? Currently this makes it so that the default installation comes with `1.15.2` instead of `1.14.16` if you check `Cargo.lock`

nickgarfield: Ah, this was unintentional. Updated!

wasm/Cargo.lock
acheroncrypto, line 1: Lock file seems to be outdated. You can use the build script in `wasm/` to update.

wasm/anchor-cli/Cargo.toml
acheroncrypto, line 1: Since this crate's version is getting updated, could you also update `anchor-lang` and `anchor-syn` version to `0.27.0` to match the current version? They are using an older git commit.

acheroncrypto (changes requested):

wasm/anchor-cli/Cargo.toml
nickgarfield: Ran the build script and updated `anchor-lang` + `anchor-syn`.

README.md
```diff
@@ -20,7 +20,7 @@
 | [num-traits](https://docs.rs/num-traits/0.2.15)                                    | 0.2.15  |
 | [pyth-sdk](https://docs.rs/pyth-sdk/0.7.0)                                         | 0.7.0   |
 | [pyth-sdk-solana](https://docs.rs/pyth-sdk-solana/0.7.0)                           | 0.7.0   |
-| [solana-program](https://docs.rs/solana-program/1.14.12)                           | 1.14.12 |
+| [solana-program](https://docs.rs/solana-program/1.14.16)                           | 1.14.16 |
```
acheroncrypto, line 23: One last thing, Playground still uses `solana-program` 1.14.12 and this PR does not update it(program crates don't live in this repository).

Can merge it after you restore this file.

PR #126, opened 2023-04-13
Make build script work on macOS and handle invalid theme key

Woody4618:
On mac os the sed command needs an extra parameter to tell it not to backup the files. 
And the themes can have an old key, which doesn't exists, so i added a fallback to default theme.


acheroncrypto (approved):
Theme issue results from another app using the same theme key in local development. Thank you for the both fixes!

PR #134, opened 2023-05-12
Anchor XORShift RNG tutorial

ZYJLiu:
Add basic Anchor example of how to use XORShift in a program utilizing the current slot as a source of pseudo-randomness.


acheroncrypto (approved):
Awesome, thank you!

PR #139, opened 2023-05-21
Added SPL token vault tutorial

Woody4618:
- Learn how to create SPL tokens with meta data
- Save tokens in an anchor program using PDAs


client/src/tutorials/SplTokenVault/files/client.ts.raw
```diff
@@ -0,0 +1,98 @@
...
+  getOrCreateAssociatedTokenAccount,
+} from "@solana/spl-token";
+import { PublicKey } from "@solana/web3.js";
+import { keypairIdentity, token, Metaplex } from "@metaplex-foundation/js";
+
+const mintAuthority = pg.wallet.keypair;
+
+const decimals = 9;
+
+let [tokenAccountOwnerPda, bump] = await PublicKey.findProgramAddress(
```
acheroncrypto, line 12: Can we use the sync version of this(deprecated)?
```suggestion
let [tokenAccountOwnerPda] =  PublicKey.findProgramAddressSync(
```

client/src/tutorials/SplTokenVault/files/client.ts.raw
```diff
@@ -0,0 +1,98 @@
...
+  [Buffer.from("token_account_owner_pda", "utf8")],
```
acheroncrypto, line 13:
```suggestion
  [Buffer.from("token_account_owner_pda")],
```

client/src/tutorials/SplTokenVault/files/client.ts.raw
```diff
@@ -0,0 +1,98 @@
...
+);
+
+console.log("tokenAccount: " + tokenAccount.address);
+console.log("TokenAccountOwnerPda: " + tokenAccountOwnerPda);
+
+let tokenAccountInfo = await getAccount(pg.connection, tokenAccount.address);
+console.log(
+  "Owned token amount: " + tokenAccountInfo.amount / BigInt(mintDecimals)
+);
+let [token_vault, bump2] = await PublicKey.findProgramAddress(
```
acheroncrypto, line 62: Same here, also better to keep variable names camelCase for consistency.
```suggestion
let [tokenVault] = PublicKey.findProgramAddressSync(
```

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        Ok(())
+    }
+
+    pub fn transfer_in(ctx: Context<TransferAccounts>, amount: u64) -> Result<()> {
+        msg!("Token amount transfer in: {}!", amount);
+        // See Tutorial page 6
+        Ok(())
+    }
+
+    pub fn transfer_out(ctx: Context<TransferAccounts>, amount: u64) -> Result<()> {
```
acheroncrypto, line 21: Warning.

```suggestion
    pub fn transfer_out(ctx: Context<TransferAccounts>, amount: u64) -> Result<()> {
```

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+            seeds=[b"token_account_owner_pda".as_ref()],
+            bump,
+            space = 8
+        )]
+    token_account_owner_pda: AccountInfo<'info>,
+
+    #[account(
+            init_if_needed,
+            payer = signer,
+            seeds=[b"token_vault".as_ref(), mint_of_token_being_sent.key().as_ref()],
```
acheroncrypto, line 43:
```suggestion
            seeds=[b"token_vault", mint_of_token_being_sent.key().as_ref()],
```

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    system_program: Program<'info, System>,
+    token_program: Program<'info, Token>,
+    rent: Sysvar<'info, Rent>,
+}
+
+#[derive(Accounts)]
+pub struct TransferAccounts<'info> {
+    // Derived PDAs
+    #[account(mut,
+            seeds=[b"token_account_owner_pda".as_ref()],
```
acheroncrypto, line 63:
```suggestion
            seeds=[b"token_account_owner_pda"],
```

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+            bump
+        )]
+    token_account_owner_pda: AccountInfo<'info>,
+
+    #[account(mut,
+            seeds=[b"token_vault".as_ref(), mint_of_token_being_sent.key().as_ref()],
```
acheroncrypto, line 69:
```suggestion
            seeds=[b"token_vault", mint_of_token_being_sent.key().as_ref()],
```

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        msg!("Token amount transfer out: {}!", amount);
+        // See Tutorial page 7
+        Ok(())
+    }
+}
+
+#[derive(Accounts)]
+pub struct Initialize<'info> {
+    // Derived PDAs
+    #[account(
```
acheroncrypto, line 31: Inside is extra indented.

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+            init_if_needed,
+            payer = signer,
+            seeds=[b"token_account_owner_pda".as_ref()],
+            bump,
+            space = 8
+        )]
+    token_account_owner_pda: AccountInfo<'info>,
+
+    #[account(
```
acheroncrypto, line 40: Inside is extra indented.

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    signer: Signer<'info>,
+    system_program: Program<'info, System>,
+    token_program: Program<'info, Token>,
+    rent: Sysvar<'info, Rent>,
+}
+
+#[derive(Accounts)]
+pub struct TransferAccounts<'info> {
+    // Derived PDAs
+    #[account(mut,
```
acheroncrypto, line 62: Inside is extra indented.

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+            seeds=[b"token_account_owner_pda".as_ref()],
+            bump
+        )]
+    token_account_owner_pda: AccountInfo<'info>,
+
+    #[account(mut,
```
acheroncrypto, line 68: Inside is extra indented.

client/src/tutorials/SplTokenVault/files/lib.rs
```diff
@@ -0,0 +1,86 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token::{Mint, Token, TokenAccount, Transfer};
```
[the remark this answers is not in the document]
Woody4618: These warnings come from the code that will be pasted in during the tutorial. Maybe it would be better to have all the code in the files already? It maybe confusing to people if the have to add imports and change _ctx to ctx etc. 
So either leave the warning or already put the whole code it. I tend to having all the code in because it makes it easier to work with and an easy reference when you want to look things up.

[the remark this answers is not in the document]
Woody4618: I removed the warning and added to the tutorial text to add transfer to the imports.

Woody4618:
All done 👍

client/src/tutorials/SplTokenVault/files/lib.rs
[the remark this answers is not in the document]
acheroncrypto: Ah sorry, I thought it was just a forgotten import.

client/src/tutorials/SplTokenVault/files/client.ts.raw
```diff
@@ -0,0 +1,110 @@
...
+  getOrCreateAssociatedTokenAccount,
+} from "@solana/spl-token";
+import { PublicKey } from "@solana/web3.js";
+import { keypairIdentity, token, Metaplex } from "@metaplex-foundation/js";
+
+const mintAuthority = pg.wallet.keypair;
+
+const decimals = 9;
+
+let [tokenAccountOwnerPda] = await PublicKey.findProgramAddressSync(
```
acheroncrypto, line 12: Using sync version.

```suggestion
let [tokenAccountOwnerPda] = PublicKey.findProgramAddressSync(
```

client/src/tutorials/SplTokenVault/files/client.ts.raw
```diff
@@ -0,0 +1,110 @@
...
+);
+
+console.log("tokenAccount: " + tokenAccount.address);
+console.log("TokenAccountOwnerPda: " + tokenAccountOwnerPda);
+
+let tokenAccountInfo = await getAccount(pg.connection, tokenAccount.address);
+console.log(
+  "Owned token amount: " + tokenAccountInfo.amount / BigInt(mintDecimals)
+);
+let [tokenVault] = await PublicKey.findProgramAddressSync(
```
acheroncrypto, line 62:
```suggestion
let [tokenVault] = PublicKey.findProgramAddressSync(
```

acheroncrypto (commented):
In review, I'm only picking the ones in the files but the markdown part also needs to be updated after making changes to program files or tests. This is also why the *Hello World* tutorials only show the code in markdown files so that it doesn't need to be synced.

There is also another advantage of using the tutorial files and let the users write it for themselves which is if a user started a tutorial, the markdown files can be updated anytime and the user will see the changes but the actual program files will not get updated since its just a file that is stored in the user's browser and they are free to do anything they want with it.

I'll think of ways to mitigate this problem, maybe a button to show the completed version of the tutorial works better?

Woody4618:
Updated

Woody4618:
I also updated this example to the newest version. Did I do that correct with that assets submodule? I just added the icon there, i hope that is enough?

acheroncrypto (commented):
> I also updated this example to the newest version. Did I do that correct with that assets submodule? I just added the icon there, i hope that is enough?

Thanks, please see https://github.com/solana-playground/assets/pull/1#pullrequestreview-1813516898. The PR part is correct but I think you've also updated the submodule to your fork in this PR which shouldn't happen. Instead, run:

```sh
git submodule update --remote
```

**after** [the asset PR](https://github.com/solana-playground/assets/pull/1) is merged.

acheroncrypto (changes requested):
Thanks, [the asset PR](https://github.com/solana-playground/assets/pull/1) is merged, only need to run:

```sh
git submodule update --remote
```

There is also a discrepancy between the actual code files and the Markdown explanation files. We need to make the changes from both places currently.

Woody4618:
Updated code and public submodule using git submodule update --remote

acheroncrypto (commented):
> Updated code and public submodule using git submodule update --remote

Thanks, it's correct now.

However, there is still a mismatch between the code files and the Markdown files. I only point them in one place in my reviews but its counterpart should also be updated. I can comment them here if you'd like but it would be much easier if you compared it in your editor.

Woody4618:
I think i changed all the differences between md files and code now. Sorry for the hassle

acheroncrypto (approved):
Thank you, this will most likely go live at the end of the month.

PR #142, opened 2023-06-19
Upgrade Solana and Anchor

Aursen:


README.md
acheroncrypto, line 1: The crates specified here are being used to build programs in playground(not the wasm crates like `solana-cli-wasm`).

Unfortunately we are not able to update the program crate versions yet because some of the dependencies still depend on Anchor `0.27.0`.

wasm/solana-cli/Cargo.toml
```diff
@@ -29,9 +29,9 @@ solana-cli-output-wasm = { path = "../utils/solana-cli-output" }
 solana-client-wasm = { path = "../solana-client" }
 solana-extra-wasm = { path = "../utils/solana-extra" }
 solana-playground-utils-wasm = { path = "../utils/solana-playground-utils" }
-solana-remote-wallet = { version = "=1.14.16", default-features = false }
+solana-remote-wallet = { version = ">=1.14, <1.17", default-features = false }
 solana-sdk = "*"
-solana-version = "=1.14.16"
```
acheroncrypto, line 34: Specifying dependencies to be as general as possible is helpful for libraries like `anchor-lang` where a lot of people depend on it but currently `solana-cli-wasm` is only being used to generate a WASM binary(no code depends on it), which means we can, and should be, as specific as possible with our dependencies.

wasm/solana-client/Cargo.toml
```diff
@@ -22,8 +22,8 @@ reqwest = { version = "0.11.9", features = ["json"] }
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
-solana-extra-wasm = { path = "../utils/solana-extra", version = "=1.14.16" }
-solana-sdk = "=1.14.16"
+solana-extra-wasm = { path = "../utils/solana-extra", version = ">=1.14, <1.17" }
```
acheroncrypto, line 25: This is a local dependency and I think both `solana-client-wasm` and `solana-extra-wasm` should continue follow the same versioning as the official Solana crates.

acheroncrypto (changes requested):

Aursen:
okay I understand the architecture better, I'll just need the wasm client to have anchor compile under wasm

acheroncrypto:
> okay I understand the architecture better, I'll just need the wasm client to have anchor compile under wasm

It should be able to compile because `0.28.0` works with both `1.14` and `1.16`.

Aursen:
> > okay I understand the architecture better, I'll just need the wasm client to have anchor compile under wasm
> 
> It should be able to compile because `0.28.0` works with both `1.14` and `1.16`.

It will not compile because the solana 1.14 client does not have `UnexpectedMessageError`

PR #147, opened 2023-07-29
comments fix

tanim0la:
Fixed comment in; 

TinyAdventure: in the `Initialize` account struct

(u8 = 8 bits unsigned integer = 8 bytes)  changed to (u8 = 8 bits unsigned integer = 1 byte) 

# 

TinyAdventureTwo: in the `InitializeLevelOne` account struct

(u8 = 8 bits unsigned integer = 8 bytes)  changed to (u8 = 8 bits unsigned integer = 1 byte)


acheroncrypto (approved):
Correct, thank you.

PR #149, opened 2023-08-08
Upgrade Solana deps to 1.16.7

Aursen:


acheroncrypto:
General thought process has been to only bump for minor versions e.g. `1.16`, `1.17`... unless we make changes to the library API.

This is mainly due to keep the commit history clean for the monorepo. If needed, we might separate the WASM client from the monorepo and automate releases based on Solana releases but it doesn't look like it's needed right now since the library itself doesn't change frequently.

The versioning mainly follows Solana's versioning because of the dependence to `solana-sdk` types(to avoid potential conflicts).

Aursen:
I made this PR because it fixed a few problems with dependencies. I think automating releases sounds like a good idea, especially when there's a wasm client for Anchor. So I'm closing this PR

PR #162, opened 2023-11-07
feature: allow bare imports in clients

steveluscher:
# Summary

This regex presupposes that imports will have specifiers, or whatever they're called. This PR aims to enable ‘bare’ imports like this:

```ts
import "foo";
```

The next PR in this stack (#163) makes use of this.


acheroncrypto (approved):
Thanks!

PR #164, opened 2023-11-08
Parallelize awaits

steveluscher:
Each commit in this PR rolls up serial awaits and await-in-a-loops into a `Promise.all()`

# Test Plan

I didn't _really_ test this rigorously.

I did however take some before/after Lighthouse scores loading the playground in a browser.

| Before | After |
|---|---|
|image|image|

### DIFF
--- a/client/src/commands/sugar/commands/deploy/collection.ts
+++ b/client/src/commands/sugar/commands/deploy/collection.ts
@@ -33,43 +33,42 @@ export const createCollection = async (
 
   const payer = metaplex.identity().publicKey;
 
-  const [
-    blockhashInfo,
-    [createAtaIx, mintToIx],
-    createMintAccountIx,
-    initMintIx,
-  ] = await Promise.all([
-    // Fetch the latest blockhash to use to specify the transaction lifetime
-    metaplex.connection.getLatestBlockhash(),
-    // Get the [createAta, mintTo] instructions
-    (async () => {
-      const ataPk = await getAssociatedTokenAddress(collectionMintPk, payer);
-      return await Promise.all([
-        // Create associated account
-        createAssociatedTokenAccountInstruction(
-          payer,
-          ataPk,
-          payer,
-          collectionMintPk
-        ),
-        // Mint
-        createMintToInstruction(collectionMintPk, ataPk, payer, 1),
-      ]);
-    })(),
-    // Create mint account
-    (async () =>
-      SystemProgram.createAccount({
-        fromPubkey: payer,
-        lamports: await metaplex.connection.getMinimumBalanceForRentExemption(
-          MINT_SIZE
-        ),
-        newAccountPubkey: collectionMintPk,
-        programId: TOKEN_PROGRAM_ID,
-        space: MINT_SIZE,
-      }))(),
-    // Initialize mint
-    createInitializeMintInstruction(collectionMintPk, 0, payer, payer),
-  ]);
+  // Create mint account
+  const createMintAccountIx = await SystemProgram.createAccount({
+    fromPubkey: payer,
+    lamports: await metaplex.connection.getMinimumBalanceForRentExemption(
+      MINT_SIZE
+    ),
+    newAccountPubkey: collectionMintPk,
+    programId: TOKEN_PROGRAM_ID,
+    space: MINT_SIZE,
+  });
+
+  // Initialize mint
+  const initMintIx = await createInitializeMintInstruction(
+    collectionMintPk,
+    0,
+    payer,
+    payer
+  );
+
+  const ataPk = await getAssociatedTokenAddress(collectionMintPk, payer);
+
+  // Create associated account
+  const createAtaIx = await createAssociatedTokenAccountInstruction(
+    payer,
+    ataPk,
+    payer,
+    collectionMintPk
+  );
+
+  // Mint
+  const mintToIx = await createMintToInstruction(
+    collectionMintPk,
+    ataPk,
+    payer,
+    1
+  );
 
   const creator: Creator = {
     address: payer,
@@ -136,6 +135,8 @@ export const createCollection = async (
     ]
   );
   tx.feePayer = payer;
+
+  const blockhashInfo = await metaplex.connection.getLatestBlockhash();
   tx.recentBlockhash = blockhashInfo.blockhash;
 
   await metaplex
--- a/client/src/commands/sugar/commands/deploy/collection.ts
+++ b/client/src/commands/sugar/commands/deploy/collection.ts
@@ -33,42 +33,43 @@ export const createCollection = async (
 
   const payer = metaplex.identity().publicKey;
 
-  // Create mint account
-  const createMintAccountIx = await SystemProgram.createAccount({
-    fromPubkey: payer,
-    lamports: await metaplex.connection.getMinimumBalanceForRentExemption(
-      MINT_SIZE
-    ),
-    newAccountPubkey: collectionMintPk,
-    programId: TOKEN_PROGRAM_ID,
-    space: MINT_SIZE,
-  });
-
-  // Initialize mint
-  const initMintIx = await createInitializeMintInstruction(
-    collectionMintPk,
-    0,
-    payer,
-    payer
-  );
-
-  const ataPk = await getAssociatedTokenAddress(collectionMintPk, payer);
-
-  // Create associated account
-  const createAtaIx = await createAssociatedTokenAccountInstruction(
-    payer,
-    ataPk,
-    payer,
-    collectionMintPk
-  );
-
-  // Mint
-  const mintToIx = await createMintToInstruction(
-    collectionMintPk,
-    ataPk,
-    payer,
-    1
-  );
+  const [
+    blockhashInfo,
+    [createAtaIx, mintToIx],
+    createMintAccountIx,
+    initMintIx,
+  ] = await Promise.all([
+    // Fetch the latest blockhash to use to specify the transaction lifetime
+    metaplex.connection.getLatestBlockhash(),
+    // Get the [createAta, mintTo] instructions
+    (async () => {
+      const ataPk = await getAssociatedTokenAddress(collectionMintPk, payer);
+      return await Promise.all([
+        // Create associated account
+        createAssociatedTokenAccountInstruction(
+          payer,
+          ataPk,
+          payer,
+          collectionMintPk
+        ),
+        // Mint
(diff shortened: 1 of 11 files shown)


client/src/commands/sugar/commands/deploy/collection.ts
```diff
@@ -33,42 +33,43 @@ export const createCollection = async (
...
+        lamports: await metaplex.connection.getMinimumBalanceForRentExemption(
+          MINT_SIZE
+        ),
+        newAccountPubkey: collectionMintPk,
+        programId: TOKEN_PROGRAM_ID,
+        space: MINT_SIZE,
+      }))(),
+    // Initialize mint
+    createInitializeMintInstruction(collectionMintPk, 0, payer, payer),
+  ]);
```
acheroncrypto, line 72: I don't think these functions are asynchronous anymore. At the time of writing they were using `PublicKey.findProgramAddress` which basically forced entire applications to be written in async code 😅.

client/src/components/EditorWithTabs/Editor/Monaco/languages/rust/rust-analyzer/rust-analyzer.ts
```diff
@@ -158,16 +160,20 @@ const loadLocalFiles = async () => {
...
+
+      if (new RegExp(`${crate}::`, "gm").test(model.getValue())) {
+        return loadDependency(crate);
+      } else if (status !== "empty") {
+        return loadDependency(crate, { empty: true });
+      } else {
+        return null;
+      }
+    })
+  );
```
acheroncrypto, line 176: For a simple benchmark, I've added `console.time("update");` [before here](https://github.com/solana-playground/solana-playground/blob/36bbb1a7e8db4760ef49a75d5a7cb5995192ce86/client/src/components/EditorWithTabs/Editor/Monaco/languages/rust/rust-analyzer/rust-analyzer.ts#L161) and `console.timeEnd("update");` [after here](https://github.com/solana-playground/solana-playground/blob/36bbb1a7e8db4760ef49a75d5a7cb5995192ce86/client/src/components/EditorWithTabs/Editor/Monaco/languages/rust/rust-analyzer/rust-analyzer.ts#L170).

**Previous:**

```
rust-analyzer.ts:172 update: 96.833984375 ms
rust-analyzer.ts:172 update: 0.086181640625 ms
rust-analyzer.ts:172 update: 0.093017578125 ms
rust-analyzer.ts:172 update: 0.406982421875 ms
rust-analyzer.ts:172 update: 0.31884765625 ms
rust-analyzer.ts:172 update: 0.06298828125 ms
rust-analyzer.ts:172 update: 0.067138671875 ms
rust-analyzer.ts:172 update: 0.35595703125 ms
rust-analyzer.ts:172 update: 0.092041015625 ms
rust-analyzer.ts:172 update: 0.06494140625 ms
rust-analyzer.ts:172 update: 0.130859375 ms
rust-analyzer.ts:172 update: 0.049072265625 ms
```


**After:**

```
rust-analyzer.ts:178 update: 49.904052734375 ms
rust-analyzer.ts:178 update: 0.131103515625 ms
rust-analyzer.ts:178 update: 5.697021484375 ms
rust-analyzer.ts:178 update: 19.4228515625 ms
rust-analyzer.ts:178 update: 7.8291015625 ms
rust-analyzer.ts:178 update: 15.10205078125 ms
rust-analyzer.ts:178 update: 12.961181640625 ms
rust-analyzer.ts:178 update: 6.27099609375 ms
rust-analyzer.ts:178 update: 2.1630859375 ms
rust-analyzer.ts:178 update: 7.06494140625 ms
rust-analyzer.ts:178 update: 7.393798828125 ms
rust-analyzer.ts:178 update: 7.012939453125 ms
```

While the first load is about 2x faster, each update is a lot slower compared to previous. `update` is called each time the content changes and I've tested with keystrokes. We could throttle the update but I didn't see a reason to do it because each update was already sub 1ms. I'm not sure why this is a lot slower but this is why optimimizing without benchmarking could be dangerous.

Please let me know if you have different results.

> **Note**
> The performance of [rust-analyzer](https://github.com/solana-playground/solana-playground/tree/d8f1696627ad7b70fe60c33c2c6c87c451296562/wasm/rust-analyzer) is not limited by these calls as the time it takes to load the crates is usually less than 1% of Rust Analyzer's initialization duration.

acheroncrypto (commented):
Thank you for taking to time to write this PR. After https://github.com/solana-playground/solana-playground/pull/163#issuecomment-1802307007 I've realized [declarations file](https://github.com/solana-playground/solana-playground/blob/e641980d5ac2bdd00caf142509e9c34bf2dbd347/client/src/components/EditorWithTabs/Editor/Monaco/languages/typescript/declarations/declarations.ts) should be asynchronously imported and the imports inside should be made sync, I guess you had similar intuition 😅

We actually use `Promise.all` in places where there are a lot of requests, for example:

- https://github.com/solana-playground/solana-playground/blob/36bbb1a7e8db4760ef49a75d5a7cb5995192ce86/client/src/utils/bpf-upgradeable-browser/bpf-upgradeable-browser.ts#L473
- https://github.com/solana-playground/solana-playground/blob/36bbb1a7e8db4760ef49a75d5a7cb5995192ce86/client/src/commands/sugar/commands/mint/mint.ts#L80
- https://github.com/solana-playground/solana-playground/blob/36bbb1a7e8db4760ef49a75d5a7cb5995192ce86/client/src/commands/sugar/commands/upload/upload.ts#L171

This is my personal opinion, I'm a fan of changes like https://github.com/solana-playground/solana-playground/pull/164/commits/54327f7cdd281650f405880bcabfcc3475791762 but not https://github.com/solana-playground/solana-playground/pull/164/commits/783e319ce0964a9d886986acccc2108eb88f758e. I think if a code change makes it harder to make sense of the code, it should give noticable amount of performance benefits. In the case of the second one, I don't think there are noticable performance improvements but it makes the code a lot more complicated(at least for my sequential brain).

> **Warning**
> This way also ended up being much slower for Rust Analyzer(please see comment).

I'm also having a hard time understanding how these changes could affect Lighthouse(5.1s -> 4.7s) because only 1/6 of them can run(either `rust-analyzer` or `ts-worker`) in the default page(with editor open) but they only start executing after everything else is painted. Please enlighten me if I'm missing something here.

### TL;DR;

I'm not a fan of optimizing JS code unless there are noticable performance issues.

client/src/commands/sugar/commands/deploy/collection.ts
steveluscher: Oh, we're bringin' 'em back, baby. I'm glad you didn't change these callsites, because the new web3.js is making these async again.

Reason: the `SubtleCrypto` digest method is async.

client/src/components/EditorWithTabs/Editor/Monaco/languages/rust/rust-analyzer/rust-analyzer.ts
steveluscher: Oh for sure. This is because you were able to synchronously escape the update if all the crates were `full`, and I made it so that you unconditionally wait until the next runloop.

Fixed in ea8c7eda0e597b86b106a0b12127cfa444bd5fcb.

steveluscher:
> This is my personal opinion, I'm a fan of changes like https://github.com/solana-playground/solana-playground/commit/54327f7cdd281650f405880bcabfcc3475791762 but not https://github.com/solana-playground/solana-playground/commit/783e319ce0964a9d886986acccc2108eb88f758e

Yeah, JavaScript is garbage. This is why we desperately need `concurrent { }` blocks [like in Hack](https://docs.hhvm.com/hack/asynchronous-operations/concurrent).

```
concurrent {
  // Create mint account
  async {
    const lamports = await metaplex.connection.getMinimumBalanceForRentExemption(
      MINT_SIZE
    );
    const createMintAccountIx = await SystemProgram.createAccount({
      fromPubkey: payer,
      lamports,
      newAccountPubkey: collectionMintPk,
      programId: TOKEN_PROGRAM_ID,
      space: MINT_SIZE,
    });
  }

  // Initialize mint
  const initMintIx = await createInitializeMintInstruction(
    collectionMintPk,
    0,
    payer,
    payer
  );

  async {
    const ataPk = await getAssociatedTokenAddress(collectionMintPk, payer);
    // Create associated account
    const createAtaIx = await createAssociatedTokenAccountInstruction(
      payer,
      ataPk,
      payer,
      collectionMintPk
    );
  }
}
```

Perfection.

> I'm also having a hard time understanding how these changes could affect Lighthouse…

No idea either.

Anyway; please do close this, unmerged! I was just procrastinating because I saw a serial await and…

image

client/src/commands/sugar/commands/deploy/collection.ts
acheroncrypto: Almost everything in JS seems to be eventually written in async because that one little async call somewhere 😅

acheroncrypto:
> Yeah, JavaScript is garbage. This is why we desperately need `concurrent { }` blocks [like in Hack](https://docs.hhvm.com/hack/asynchronous-operations/concurrent).

That would be amazing!

> Anyway; please do close this, unmerged! I was just procrastinating because I saw a serial await and…
> 
> image

You are not alone in this 😄

client/src/commands/sugar/commands/deploy/collection.ts
acheroncrypto: Reverted this change in https://github.com/solana-playground/solana-playground/pull/164/commits/1a4c7f750908c86e82dc5729a081fafc8e1b5bd2 — others look great

client/src/components/EditorWithTabs/Editor/Monaco/languages/typescript/declarations/importable.ts
acheroncrypto, line 1: This will be very useful with transitive deps because we were only making 1 fetch per package but now it's a lot more than 1. I will add some caching and maybe allow max depth of 1 when including transitive deps because importing `@metaplex-foundation/js` causes about 1500 requests 😅

acheroncrypto (approved):
> Anyway; please do close this, unmerged! I was just procrastinating because I saw a serial await and…

No need — it was just https://github.com/solana-playground/solana-playground/pull/164/commits/783e319ce0964a9d886986acccc2108eb88f758e and https://github.com/solana-playground/solana-playground/pull/164/commits/43a920f00b9f0c47faeeddd85dc6f8d7ac032b21 that I didn't love, and wasn't sure if `rust-analyzer` would work properly. All seem to work great and it will especially help transitive dependencies so it's a merge!

PR #185, opened 2024-01-19
Update to seahorse-dev 0.2.0

mcintyre94:
- Update Playground to the latest version of Seahorse
- Most importantly, includes https://github.com/solana-developers/seahorse/pull/10 which fixes Seahorse for Anchor 0.29.0

### DIFF
--- a/client/src/frameworks/seahorse/files/src/fizzbuzz.py
+++ b/client/src/frameworks/seahorse/files/src/fizzbuzz.py
@@ -1,5 +1,5 @@
 # fizzbuzz
-# Built with Seahorse v0.1.1
+# Built with Seahorse v0.2.0
 #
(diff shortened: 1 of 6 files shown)


acheroncrypto (commented):
Not too important but would be nice to also update the version in default project comments:

https://github.com/solana-playground/solana-playground/blob/94ba05a53f8b4a027dbe251032f2dd185689683d/client/src/frameworks/seahorse/files/src/fizzbuzz.py#L2

https://github.com/solana-playground/solana-playground/blob/94ba05a53f8b4a027dbe251032f2dd185689683d/vscode/src/commands/create/seahorse.ts#L16

acheroncrypto (approved):
Thanks!

Issue #187, opened 2024-01-24
Error when testing with seeds


lostintime101:
![err](https://github.com/solana-playground/solana-playground/assets/92709487/cbee63c6-ed75-4de2-bb28-0d68fd3678d7)

https://github.com/solana-playground/solana-playground/assets/92709487/3d1b718f-8ee4-4e42-8c05-d2e386d425eb

Same error across separate devices/OS: Linux Firefox & Windows Chrome

acheroncrypto:
The website was updated an hour ago so this was a new bug, it should now be fixed with 59d36a82580b8f4a31fdfc6b259515b3259354f0. Please refresh the page to fetch the newest version.

Thanks for creating the issue, especially with a video example.

lostintime101:
It's fixed. Great job!

PR #189, opened 2024-01-25
Allow pasting into fields in test UI

### LINKED ISSUE #188: Unable to paste into public key field in test UI
In a field that is expecting a public key, pasting a public key (in this case  I've copied`fizzbuzz` from init `DPskm...`) doesn't work. The input is immediately cleared again.

Screenshot 2024-01-25 at 08 56 43

mcintyre94:
I'm not 100% about this fix, I might be missing some nuance about how search bar items work

But the reason pasting was not working is because `item.value ||= clipboardText` was never updating, because `item.value` was truthy:
Screenshot 2024-01-25 at 10 13 24

This PR changes it to always replace `item.value` with the pasted text

~~An alternative fix would be to remove the `{current: true}` from the Custom item, so that `item.value` is just whatever that field contains, but I'm not sure what that represents/would do.~~
Actually I guess that wouldn't work because then you couldn't paste unless the field was empty. I'm not sure if there are circumstances where we shouldn't change `item.value` on paste though. 

Fixes #188

### DIFF
--- a/client/src/components/SearchBar/SearchBar.tsx
+++ b/client/src/components/SearchBar/SearchBar.tsx
@@ -341,7 +341,7 @@ const SearchBar: FC<SearchBarProps> = ({
             if (item) {
               // `item.value` can be an empty string because it's not updated as
               // soon as `onPaste` executes.
-              if (typeof item === "object") item.value = clipboardText;
+              if (typeof item === "object") item.value ||= clipboardText;
 
               // Include a timeout otherwise the pasted text is appended to the
               // value we set instead of overriding it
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
https://github.com/solana-playground/solana-playground/pull/189/commits/20a935c2e5f8a3ff2bd5d590e3decbd61ecca775 would override any existing text in the input. I think the problem is caused by the [last commit](https://github.com/solana-playground/solana-playground/commit/59d36a82580b8f4a31fdfc6b259515b3259354f0) that fixed the issue https://github.com/solana-playground/solana-playground/issues/187.

Specifically the line

https://github.com/solana-playground/solana-playground/blob/59d36a82580b8f4a31fdfc6b259515b3259354f0/client/src/components/SearchBar/SearchBar.tsx#L103

We're setting the value to `props.value` which is an older value, in this case resets the input. I've changed it to use `inputRef.current?.value ?? props.value`. Let me know if this fixes it.

mcintyre94:
Gotcha thanks, fix looks good! :)

PR #204, opened 2024-02-19
wss if https

HardhatChad:
Use secure websocket connection if RPC url is secure


acheroncrypto (approved):
Thanks! Do you need a release with this fix?

HardhatChad:
Whenever is convenient. I can use a patch for now

PR #247, opened 2024-06-14
Update Playground to Anchor 0.30.0

ZYJLiu:
Here's an attempt to update Playground to 0.30.0. Please lmk what you think should be changed/updated.

- Pg server updated to build with Anchor 0.30.0 and return the new IDL format.
- UI updated to work with `declare_program` macro, and an option to add an IDL folder for external IDL files when in an Anchor project
- UI Test panel (UI to interact with Instructions/Accounts) updated to work with new IDL. Although it seems to function, not sure if there are still parts that might be broken. Not entirely sure if I've updated the files under `/client/src/utils/pg/program-interaction` correctly. Specifically had trouble figuring out the correct type for typescript.

https://github.com/solana-playground/solana-playground/assets/75003086/5ad248de-b67c-40e4-9159-5fe80163b59a

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -17,7 +17,7 @@
     "@codemirror/search": "^6.2.0",
     "@codemirror/state": "^6.1.1",
     "@codemirror/view": "^6.2.3",
-    "@coral-xyz/anchor": "^0.29.0",
+    "@coral-xyz/anchor": "^0.30.0",
     "@dnd-kit/core": "^6.0.8",
     "@dnd-kit/sortable": "^7.0.2",
     "@isomorphic-git/lightning-fs": "^4.6.0",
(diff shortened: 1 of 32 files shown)


jacobcreech:
@acheroncrypto any idea on the build issue? Seems like it just exited.

acheroncrypto:
> @acheroncrypto any idea on the build issue? Seems like it just exited.

It means there are uncommitted changes.

ZYJLiu:
any suggestions how to fix? there doesn't seem to be any uncommited changes on my local branch

acheroncrypto:
You can run `git status` before the failure happens here:

https://github.com/solana-playground/solana-playground/blob/d9b73a4b6938db7d80a34ffe1ed5ca4126b66e69/.github/workflows/ci.yml#L129-L131

to see what needs to be updated and committed.

ZYJLiu:
When I ran `git status --porcelain | wc -l | grep -w "0"` locally, the output is `0`

Johnson1s (approved):

PR #264, opened 2024-07-29
Update web3.js version + add hint to web faucet

Woody4618:
I updated the web3js version to ^1.92.1 so that the transaction logs are shown in an error case which makes debugging programs much easier. In case of a failed simulation it looks like this now: 
image

Then i added a hint to the solana foundation web faucet so that people dont get blocked in case the automatic airdrop fails. This looks like this an has a link which automatically fills in the playground wallet address into the link: 
https://faucet.solana.com/?walletAddress=7AFZJhEKbJ2awvTyZ4CZNRCGxey3kE325e1gFwA2KhQq&amount=5

image

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -40,7 +40,7 @@
     "@solana/spl-token-metadata": "^0.1.2",
     "@solana/wallet-adapter-base": "^0.9.22",
     "@solana/wallet-adapter-react": "^0.15.32",
-    "@solana/web3.js": "*",
+    "@solana/web3.js": "^1.92.1",
     "assert": "^2.0.0",
     "crypto-browserify": "^3.12.0",
     "jszip": "^3.10.1",
--- a/client/package.json
+++ b/client/package.json
@@ -40,7 +40,7 @@
     "@solana/spl-token-metadata": "^0.1.2",
     "@solana/wallet-adapter-base": "^0.9.22",
     "@solana/wallet-adapter-react": "^0.15.32",
-    "@solana/web3.js": "^1.92.1",
+    "@solana/web3.js": "*",
     "assert": "^2.0.0",
     "crypto-browserify": "^3.12.0",
     "jszip": "^3.10.1",
--- a/client/yarn.lock
+++ b/client/yarn.lock
@@ -1108,10 +1108,10 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.17.2", "@babel/runtime@^7.22.6", "@babel/runtime@^7.23.2":
-  version "7.23.7"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.23.7.tgz#dd7c88deeb218a0f8bd34d5db1aa242e0f203193"
-  integrity sha512-w06OXVOFso7LcbzMiDGt+3X7Rh7Ho8MmgPoWU3rarH+8upf+wSU/grlGbWzQyr3DkdN6ZeuMFjpdwW0Q+HxobA==
+"@babel/runtime@^7.25.0":
+  version "7.25.4"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.25.4.tgz#6ef37d678428306e7d75f054d5b1bdb8cf8aa8ee"
+  integrity sha512-DSgLeL/FNcpXuzav5wfYvHCGvynXkJbn3Zvc3823AEe9nPwW9IK4UoCSS5yGymmQzN0pCPvivtgS6/8U2kkm1w==
   dependencies:
     regenerator-runtime "^0.14.0"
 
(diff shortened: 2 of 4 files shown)


Woody4618:
@acheroncrypto when you got time could you have a look why this fails? I don't understand the error. Maybe there is something that needs to be changed in the lock file to support the new web3js version?

acheroncrypto:
> @acheroncrypto when you got time could you have a look why this fails? I don't understand the error. Maybe there is something that needs to be changed in the lock file to support the new web3js version?

Yeah, that [`wasm-bindgen-error`](https://github.com/solana-playground/solana-playground/actions/runs/10143498488/job/28045093140?pr=264#step:9:110) in CI is unrelated, and it was fixed in https://github.com/solana-playground/solana-playground/commit/3586c176bfe1b363b2587448d18fd0ac30db682c. Could you rebase this branch?

Woody4618:
Ohh I see, thanks!

Woody4618:
What do you think of this PR, makes sense? Anything you would like to see changed? :)

client/package.json
```diff
@@ -40,7 +40,7 @@
     "@solana/spl-token-metadata": "^0.1.2",
     "@solana/wallet-adapter-base": "^0.9.22",
     "@solana/wallet-adapter-react": "^0.15.32",
-    "@solana/web3.js": "*",
+    "@solana/web3.js": "^1.92.1",
```
acheroncrypto, line 43: Can we keep the version as is here and only make changes to the lock file?
```suggestion
    "@solana/web3.js": "*",
```

client/yarn.lock
```diff
@@ -3052,6 +3050,27 @@
     rpc-websockets "^7.5.0"
     superstruct "^0.14.2"
 
+"@solana/web3.js@^1.92.1":
```
acheroncrypto, line 3053: We have quite many versions of `@solana/web3.js` in our lock file:

https://github.com/solana-playground/solana-playground/blob/a92a554bf03fcfafdc93b1a4502cfec54fcf3bb2/client/yarn.lock#L2970-L3053

This might require manual editing, but I think you should just be able to use the latest version for all of them to make this update work for all packages we support.

acheroncrypto (changes requested):
> What do you think of this PR, makes sense? Anything you would like to see changed? :)

I think it's very helpful since it makes it easier to debug failed transactions, but check out my comment about updating all `@solana/web3.js`.

Woody4618:
Ok so i reverted the package.json and lock file and ran 

```
sed -i '' 's/@solana\/web3.js@[^:]*:/@solana\/web3.js@^1.92.1":/g' yarn.lock
```

to update everything to 1.92.1 then 

```
yarn install 
```

then tested locally and all works fine. Is that how you imagined it?

vscode/src/commands/deploy/deploy.ts
```diff
@@ -70,7 +70,8 @@ export const processDeploy = async (
       const airdropAmount = PgConnection.getAirdropAmount();
       if (airdropAmount !== null) {
         throw new Error(
-          errMsg + `\nYou can use airdrop command to airdrop some SOL.`
+          errMsg +
+            `\nYou can use airdrop command to airdrop some SOL. \n Or you can get devnet sol from the web faucet https://faucet.solana.com/?walletAddress=${walletKp.publicKey.toString()}&amount=5 \nOr you can follow this guide to find other ways on how to get dev net sol: https://solana.com/de/developers/guides/getstarted/solana-token-airdrop-and-faucets`
```
acheroncrypto, line 74: This line is extremely long (357 characters). Could you split it up to multiple lines like [here](https://github.com/solana-playground/solana-playground/blob/c9829073999396b58d4d06e0387bad168f1ad5a6/client/src/tutorials/create.ts#L8-L11)? Maximum of 80 characters per line is `prettier`'s default.

vscode/src/commands/deploy/deploy.ts
```diff
@@ -88,7 +89,8 @@ export const processDeploy = async (
...
```
acheroncrypto, line 93: Same here

wasm/solana-cli/src/commands/wallet.rs
```diff
@@ -551,6 +551,12 @@ pub async fn process_airdrop(
             Ok(build_balance_message(current_balance, false, true))
         }
     } else {
+        PgTerminal::log_wasm(
+            &format!(
+                "In case of rate limit try the web faucet https://faucet.solana.com/?walletAddress={:?}&amount=5 or this guide: https://solana.com/developers/guides/getstarted/solana-token-airdrop-and-faucets",
```
acheroncrypto, line 556: Same issue but [different solution](https://github.com/coral-xyz/anchor/blob/879601e63283065efb852ac3e991db23a7b22a7a/cli/src/checks.rs#L115-L119). `rustfmt` has max 100 characters limit by default.

acheroncrypto (changes requested):
> then tested locally and all works fine. Is that how you imagined it?

Yes, all versions of `@solana/web3.js` are updated now, thanks!

Only the line length issue remains.

wasm/solana-cli/src/commands/wallet.rs
Woody4618: Hey. @acheroncrypto sorry, but im too stupid to format this correctly. My formatter settings seem to be different. How can I format the rust code like the CI does?

Woody4618:
Updated :)

alexis1120 (approved):

scutuatua-crypto (approved):

PR #271, opened 2024-08-04
Rework methods for wasm client

Aursen:
### Summary of changes

- Change the Request and Response serialization
- Remove some allocations
- Add unit tests for each method
- Add missing methods


wasm/solana-client/Cargo.toml
```diff
@@ -30,7 +30,12 @@ solana-sdk = "~1.18"
 thiserror = "1"
 wasm-bindgen = { version = "0.2.88", optional = true }
 wasm-bindgen-futures = { version = "0.4.38", optional = true }
-web-sys = { version = "0.3.65", features = ["MessageEvent", "WebSocket"], optional = true }
+web-sys = { version = "0.3.65", features = [
+    "MessageEvent",
+    "WebSocket",
+], optional = true }
```
acheroncrypto, line 36: Nit: Could you disable formatting on this file so that it doesn't format like this?

acheroncrypto (commented):
Thank you for this massive PR!

Just curious, did you check how much binary size difference this change makes? ([You can build and compare our crates](https://github.com/solana-playground/solana-playground/tree/master/wasm#build-packages))

Aursen:
> Thank you for this massive PR!
> 
> Just curious, did you check how much binary size difference this change makes? ([You can build and compare our crates](https://github.com/solana-playground/solana-playground/tree/master/wasm#build-packages))

The size is reduced for anchor-cli pkg and spl-token-cli. But not for the solana-cli

acheroncrypto:
The [clippy step in our CI](https://github.com/solana-playground/solana-playground/actions/runs/10271380152/job/28421200978?pr=271) should be fixed now with https://github.com/solana-playground/solana-playground/commit/e015bec87d27ccca34d2d5e23a6e742f03db54ff — could you rebase or merge this branch?

wasm/solana-client/Cargo.toml
```diff
@@ -31,6 +31,8 @@ thiserror = "1"
 wasm-bindgen = { version = "0.2.88", optional = true }
 wasm-bindgen-futures = { version = "0.4.38", optional = true }
 web-sys = { version = "0.3.65", features = ["MessageEvent", "WebSocket"], optional = true }
+serde_tuple = "0.5"
+serde_with = "3.9"
```
acheroncrypto, line 35: Nit: dependencies are ordered alphabetically.

wasm/utils/solana-extra/src/account_decoder/mod.rs
```diff
@@ -43,6 +43,7 @@ fn slice_data(data: &[u8], data_slice_config: Option<UiDataSliceConfig>) -> &[u8
 #[serde(rename_all = "camelCase")]
 pub struct UiAccount {
     pub lamports: u64,
+    pub space: u64,
```
acheroncrypto, line 46: `space` field looks to be optional: https://github.com/anza-xyz/agave/blob/584b072b9962d1e2531564bb6cfd209746210a91/account-decoder/src/lib.rs#L48

It would also be better if the ordering of this struct was exactly the same.

Also, maybe we're able to use the upstream `solana-account-decoder` crate now? I've always wanted to get rid of this `solana-extra-wasm` crate. It looks like dependencies are managed better now, so we might finally be able to do it (not necessary for this PR).

acheroncrypto (commented):
> The size is reduced for anchor-cli pkg and spl-token-cli. But not for the solana-cli

Interesting. Do you have exact numbers?

Aursen:
> > The size is reduced for anchor-cli pkg and spl-token-cli. But not for the solana-cli
> 
> Interesting. Do you have exact numbers?

Anchor-cli: new 1,63 Mo old 1,64Mo
Solana-cli: new 3,04 Mo old  2,91Mo
Token-cli: new 2,06 Mo old 2,21Mo
The size is approximate but gives you an idea.
By the way we add extra methods

wasm/utils/solana-extra/src/account_decoder/mod.rs
Aursen: It's string decoding so yeah option is a good idea if we don't know exactly the layout

Aursen: Let's ride off in another PR to avoid all the huge changes in one PR

Aursen:
Everything is done

acheroncrypto (approved):
Amazing work, thank you!

PR #273, opened 2024-08-19
Fix pubsub for the solana-client

Aursen:
I forgot to check the pubsub client in my last commit.
I added a quick fix with this PR


acheroncrypto (approved):
Thanks!

PR #274, opened 2024-08-19
Remove dependencies from the solana-client

Aursen:
The main goal of this PR is to remove the clap dependency of the wasm client


wasm/utils/solana-extra/Cargo.toml
```diff
@@ -14,29 +14,20 @@ readme = "README.md"
...
 borsh = "0.10"
 bs58 = "0.4"
 bytemuck = "1"
-chrono = { version = "0.4", features = ["serde"] }
-clap = "3"
-console = "0.15"
 fluvio-wasm-timer = "0.2"
-humantime = "2"
-lazy_static = "1"
-log = "0.4.17" # only for nightly builds
```
acheroncrypto, line 31: This was added to fix a `nightly` issue in https://github.com/solana-playground/solana-playground/pull/57. It probably doesn't exist anymore since you also removed the `solana-frozen-abi` dependency, but could you double check to make sure?

Aursen: You're right I will revert this one and work on it when we will update all the dependencies. Because currently I'm unable to test it

Aursen:
Build on:

- rustc 1.77.0-nightly (bf8716f1c 2023-12-24)  ✅
- rustc 1.75.0 (82e1608df 2023-12-21) ✅

wasm/utils/solana-extra/Cargo.toml
Johnson1s: - name: Codecov

acheroncrypto (approved):
Glad we're getting rid of these unnecessary dependencies. Thanks!

PR #281, opened 2024-09-11
Add timeout on the WASM client

Aursen:
## Problem
The current client has no timeout when a request is sent. This can be a problem if the request does not return a response.

## Summary of changes
Replace `reqwest` with `gloo-net` and add a timeout to the request.

### DIFF
--- a/wasm/Cargo.lock
+++ b/wasm/Cargo.lock
@@ -891,16 +891,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f7144d30dcf0fafbce74250a3963025d8d52177934239851c917d29f1df280c2"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


acheroncrypto (approved):
Looks great, thank you!

katopz:
refer to https://github.com/solana-playground/solana-playground/commit/08e0214e54fe68cdd7902021a3e8cc04ed77d27b#r151512484

image

It's look like `web-sys` vs `reqwest` issue to me more than `timeout` (i did have my own implement timeout for `reqwest`) and it seem like i stuck at `web-sys` via `provider.rs` too which is normal http request there not `timeout` related.

I think we depend on `web-sys` too much on this one which cause an issue when target native wasm.

PR #286, opened 2024-09-15
fix: print_player() in Tiny Tutorial

IaroslavMazur:
Removed the 0 & 3 cases from the function, because `print_player()` can never be called when the player's position is equal to either `0` or `3` (these cases are handled explicitly in the `move_left()` and `move_right()` functions).

### DIFF
--- a/client/src/tutorials/TinyAdventure/files/lib.rs
+++ b/client/src/tutorials/TinyAdventure/files/lib.rs
@@ -39,16 +39,10 @@ mod tiny_adventure {
 }
 
 fn print_player(player_position: u8) {
-    if player_position == 0 {
-        msg!("A Journey Begins!");
-        msg!("o.......");
-    } else if player_position == 1 {
+    if player_position == 1 {
         msg!("..o.....");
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
> Removed the 0 & 3 cases from the function, because `print_player()` can never be called when the player's position is equal to either `0` or `3` (these cases are handled explicitly in the `move_left()` and `move_right()` functions).

What if the player calls `move_left` when he is on 3 or calls `move_right` when he's on 0?

IaroslavMazur:
> > Removed the 0 & 3 cases from the function, because `print_player()` can never be called when the player's position is equal to either `0` or `3` (these cases are handled explicitly in the `move_left()` and `move_right()` functions).
> 
> What if the player calls `move_left` when he is on 3 or calls `move_right` when he's on 0?

Great point, @acheroncrypto, thank you!

Closing the PR.

therealwiggles (commented):
I like your work

therealwiggles (commented):
I like it

PR #287, opened 2024-09-15
fix: call moveRight() with the correct password

IaroslavMazur:
Not passing a password when calling the `moveRight()` function leads to the user being charged the 0.10 SOL without any way of recovering it.

### DIFF
--- a/client/src/tutorials/TinyAdventureTwo/files/client.ts.raw
+++ b/client/src/tutorials/TinyAdventureTwo/files/client.ts.raw
@@ -53,7 +53,7 @@ console.log("o........💎");
 // Here we move to the right three times and collect the chest at the end of the level
 for (let i = 0; i < 3; i++) {
   txHash = await pg.program.methods
-    .moveRight()
+    .moveRight("gib")
     .accounts({
       chestVault: chestVaultAccount,
       gameDataAccount: globalLevel1GameDataAccount,


client/src/tutorials/TinyAdventureTwo/files/client.ts.raw
```diff
@@ -53,7 +53,7 @@ console.log("o........💎");
 // Here we move to the right three times and collect the chest at the end of the level
 for (let i = 0; i < 3; i++) {
   txHash = await pg.program.methods
-    .moveRight()
+    .moveRight("gib")
```
acheroncrypto, line 56: This instruction doesn't seem to be taking any parameters in the default code:

https://github.com/solana-playground/solana-playground/blob/be4bcbe25e4364c388377b32300c3f12b8246e76/client/src/tutorials/TinyAdventureTwo/files/lib.rs#L41

The password is mentioned later on in the tutorial in page 3:

https://github.com/solana-playground/solana-playground/blob/be4bcbe25e4364c388377b32300c3f12b8246e76/client/src/tutorials/TinyAdventureTwo/pages/3.md#L20-L23

And the same page also mentions that you need to make this change from client side:

https://github.com/solana-playground/solana-playground/blob/be4bcbe25e4364c388377b32300c3f12b8246e76/client/src/tutorials/TinyAdventureTwo/pages/3.md#L79

Thefore, the default code looks to be correct and this change would break it.

IaroslavMazur: Thanks for getting back, @acheroncrypto!

It's been a long time since I opened the PR, and I no longer remember the context that has prompted me to propose the change.

Your retort sounds reasonable enough to me, so, I'll just close the PR.

PR #291, opened 2024-09-26
Add support of jupiter-cpi crate

Pan-chao:
Adding jupiter-cpi would allow users to interact with pretty much all swap functions on Solana.

Hopefully I did it right, added the crate in Cargo.toml as the documentation recommanded: https://station.jup.ag/docs/apis/cpi

Have a nice evening.

### DIFF
--- a/server/programs/Cargo.toml
+++ b/server/programs/Cargo.toml
@@ -19,6 +19,7 @@ borsh = "*"
 borsh-derive = "*"
 bytemuck = "*"
 bytemuck_derive = "*"
+jupiter-cpi = { git = "https://github.com/jup-ag/jupiter-cpi", rev = "5eb8977" }
 mpl-bubblegum = "*"
 mpl-token-auth-rules = { version = "*", features = ["no-entrypoint"] }
 mpl-token-metadata = "*"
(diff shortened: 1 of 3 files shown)


server/programs/Cargo.toml
```diff
@@ -19,6 +19,7 @@ borsh = "*"
 borsh-derive = "*"
 bytemuck = "*"
 bytemuck_derive = "*"
+jupiter-cpi = { git = "https://github.com/jup-ag/jupiter-cpi", rev = "5eb8977" }
```
acheroncrypto, line 22: The client currently only support crates.io crates, but it should be relatively easy to also support crates from `git`.

acheroncrypto (commented):
Thanks for creating this PR!

A few more places need to be updated when adding crates. However, the main issue is that we're currently planning to move away from CPI crates, and instead use [`declare_program!`](https://docs.rs/anchor-lang/0.30.1/anchor_lang/macro.declare_program.html) which will allow us to support most programs, including Jupiter, without having to add it as a dependency. We'll be able to do this after upgrading to Anchor v0.30+, but it's a bit tricky for us, and I haven't been able to find the time to do it just yet.

There is already a PR (https://github.com/solana-playground/solana-playground/pull/247) to upgrade to Anchor v0.30, but unfortunately, given the scope of the upgrade, it will take longer for me to review it and give feedback than for me to start the upgrade from scratch.

Zubairrahman11 (approved):

PR #328, opened 2025-01-17
Adding MPL Core

blockiosaurus:
Adding the MPL Core program. I excluded the TS clients for now because there's no Umi support in playground.


omgbbqhaxx:
hey please accept this request immediately..

acheroncrypto (changes requested):
Thanks. CI passes because we don't currently check for this but [the program lock file](https://github.com/solana-playground/solana-playground/blob/22f84c8fb55c30d4f7cd43eb7058ef4aeb8c98f1/server/programs/Cargo.lock) also needs to be updated.

Emadalshamery:
Let's go

PR #346, opened 2025-03-10
Fix `body stream already read` error in `solana-client-wasm`

### LINKED ISSUE #333: `solana-client-wasm` results in `body stream already read` error
### Problem

WASM CLI commands that use the `solana-client-wasm` lib results in `body stream already read` error.

For example:

```
$ solana balance
Process error: Client error: TypeError: Failed to execute 'text' on 'Response': body stream already read
```

Haven't checked when exactly this started happening, but it's probably coming from either https://github.com/solana-playground/solana-playground/pull/271 or https://github.com/solana-playground/solana-playground/pull/281.

### Solution

Debug and fix the problem.

Aursen:
Fixes #333

### DIFF
--- a/wasm/solana-client/src/client.rs
+++ b/wasm/solana-client/src/client.rs
@@ -85,7 +85,7 @@ impl WasmClient {
 
     async fn send<T: Method, R: DeserializeOwned>(&self, request: T) -> ClientResult<R> {
         let Provider::Http(provider) = &self.provider;
-        provider.send(&request).await?.result
+        Ok(provider.send(&request).await?.result)
     }
 
     pub async fn get_balance_with_commitment(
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
Nice fix, thanks!

I was just about to merge this, but I still got the same error when I ran the `solana airdrop` command:

```
$ solana airdrop 1
Requesting airdrop of 1 SOL
Process error: Client error: TypeError: Failed to execute 'text' on 'Response': body stream already read
```

The airdrop was successful but we still got the error.

All other commands I tried seems to work now except this one. It's worth noting that I only tried a few commands though.

Aursen:
Sorry for the delay but I fixed the method @acheroncrypto

acheroncrypto (approved):
Awesome, thanks!

PR #436, opened 2026-02-20
fix: skip declare_id! inside comments during build

agentmila:
## Summary

Fixes #434 — The build system injects the program ID into commented `declare_id!` macros instead of the active one.

## Problem

The regex used to locate `declare_id!` matches the first occurrence in the file regardless of whether it's inside a comment (`//` or `/* */`). When a commented `declare_id!` appears before the real one, the program ID gets injected into the comment, leaving the active macro empty and breaking the build.

## Fix

Replace the single-pass regex with line-by-line processing that tracks block comment state (`/* ... */`) and skips:
- Lines inside block comments
- Lines starting with `//` single-line comments

Only uncommented `declare_id!` macros are updated with the generated program ID.

## Changes

- **`client/src/commands/build/build.ts`** — Updated `updateIdRust()` to process lines individually with comment tracking
- **`vscode/src/commands/build/build.ts`** — Updated both `getDeclareIdFile()` detection and `updateId()` replacement to skip commented code

## Test Cases Covered

All cases from the issue:
1. Block comment with empty active macro → only active macro updated ✅
2. Single-line comment with empty active macro → only active macro updated ✅  
3. Block comment with real ID + empty active macro → comment preserved, active macro updated ✅
4. No comments (normal case) → works as before ✅

### DIFF
--- a/vscode/src/commands/build/build.ts
+++ b/vscode/src/commands/build/build.ts
@@ -39,15 +39,55 @@ export const processBuild = async () => {
   }
 
   // Change program id in file system
-  const rustDeclareIdRegex = new RegExp(
-    /^\s*(([\w_]+::)*)declare_id!\("(\w*)"\)/gm
-  );
+  const rustDeclareIdRegex = /^\s*(([\w_]+::)*)declare_id!\("(\w*)"\)/;
   const pythonDeclareIdRegex = new RegExp(/^declare_id\(("|')(\w*)("|')\)/gm);
+
+  /**
+   * Update the `declare_id!` macro in Rust source content with the current
+   * program's public key, skipping occurrences inside comments.
+   *
+   * Uses the same comment-skipping approach as the client build command.
+   */
+  const updateIdRust = (content: string) => {
+    let updated = false;
+    let insideBlockComment = false;
+    const programPkStr = programKp.publicKey.toBase58();
+    const newContent = content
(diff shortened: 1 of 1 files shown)


vscode/src/commands/build/build.ts
```diff
@@ -39,15 +39,32 @@ export const processBuild = async () => {
...
+      }
+      if (trimmed.startsWith("/*")) {
+        inBlockComment = !trimmed.includes("*/");
+        continue;
+      }
+      if (trimmed.startsWith("//")) continue;
+      if (rustDeclareIdLineRegex.test(line)) return true;
+    }
+    return false;
+  };
```
acheroncrypto: This logic seems very similar to the one below. It would be better to reuse it rather than duplicating it.

acheroncrypto (changes requested):
https://github.com/solana-playground/solana-playground/pull/435 was created before this PR and it got merged. This PR, however, also fixes the problem in VS Code. Tbh the extension has been unmaintained for a long time, but it probably wouldn't hurt to merge this (needs rebase).

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

acheroncrypto (approved):
Thanks!

PR #442, opened 2026-04-10
Pin the node version to allow informing that the engine is wrong

rogaldh:
## Description

Pin the supported Node version at the project level instead of polyfilling `navigator` for older Node runtimes.

## Problem

`yarn start` fails on Node ≤20 because `tsx` loads `xterm` / `lightning-fs`, which reads `navigator` at module init. Node 21+ exposes `navigator` globally, so the issue isn't reproducible on the README's recommended `v22.20.0`.

`.nvmrc` keeps the version to allow using the proper node version for tooling

## Solution

- Add `client/.nvmrc` (`22.20.0`) for `nvm use` auto-switching.
- Add `engines.node: ">=22.20.0"` in `client/package.json` to match the README.


rogaldh:
@acheroncrypto Please take a look

acheroncrypto (commented):
I'm not able to reproduce this problem locally while running natively. We also test the full prod builds in [CI](https://github.com/solana-playground/solana-playground/blob/679619ae753dbee137ec21dcfa882b3e0613a713/.github/workflows/ci.yml#L99) (e.g. [Build](https://github.com/solana-playground/solana-playground/actions/runs/24275973401/job/70889922620)).

Do you have [the recommended versions](https://github.com/solana-playground/solana-playground/blob/679619ae753dbee137ec21dcfa882b3e0613a713/README.md#recommended-versions)?

rogaldh:
@acheroncrypto I've checked again. It looks like I had the v20 version at the terminal, and it failed as the navigator was introduced at node>=v22. Sorry.

I changed the PR to add information that v22 is required. With these changes, yarn will inform if the version is wrong.

acheroncrypto (approved):
Thanks!

PR #445, opened 2026-04-16
Upgrade TypeScript from 4.9.5 to 5.0.4

rogaldh:
Upgrades the TypeScript compiler from 4.9.5 to 5.0.4 and adds type checking and build verification scripts.

### Problem

`@solana/codecs-data-structures@2.0.0-rc.1` (transitive dependency via `@solana/web3.js`) uses `const` type parameters — a syntax introduced in TypeScript 5.0. The TS 4.9.5 compiler cannot parse these declarations, producing 160 errors across 4 files:

```
node_modules/@solana/codecs-data-structures/dist/types/discriminated-union.d.ts:75:54
error TS1139: Type parameter declaration expected.

export declare function getDiscriminatedUnionEncoder<const TVariants extends Variants<Encoder<any>>, ...
Found 160 errors in 4 files.
```

**To reproduce:**
1. Checkout `playground/master`
2. `cd client && yarn install`
3. `yarn exec tsc -- --noEmit`

### Solution

- Upgrade `typescript` from `=4.9.5` to `=5.0.4` with a `resolutions` entry to ensure a single version across the dependency tree
- Add `"ignoreDeprecations": "5.0"` to `tsconfig.json` — the project uses `keyofStringsOnly` (`keyof` in template literal types). That was deprecated in TS 5.0 and will be removed in TS 5.5
- Add `test` script (`tsc --noEmit` + `craco build`) to verify both type correctness and build compatibility with `react-scripts`

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -54,7 +54,7 @@
     "shiki": "^0.14.4",
     "stream-browserify": "^3.0.0",
     "styled-components": "^5.3.3",
-    "typescript": "=4.9.5",
+    "typescript": "=5.0.4",
     "vscode-oniguruma": "^1.7.0",
     "vscode-textmate": "^9.0.0",
     "xterm": "^4.18.0",
@@ -77,7 +77,8 @@
   },
   "resolutions": {
     "@solana/buffer-layout": "^4.0.0",
-    "styled-components": "^5"
+    "styled-components": "^5",
+    "typescript": "5.0.4"
   },
   "engines": {
     "node": ">=22.20.0"
@@ -85,7 +86,8 @@
   "scripts": {
     "setup": "git submodule update --init && yarn build-wasm && yarn generate",
     "start": "yarn generate && craco start",
-    "build": "yarn generate && GENERATE_SOURCEMAP=false craco build",
+    "build": "yarn generate && yarn run build-craco",
+    "build-craco": "GENERATE_SOURCEMAP=false craco build",
     "build-wasm": "build=\"../wasm/build.sh\" && chmod +x $build && $build --update",
     "format": "prettier --write src/",
     "check-format": "prettier --check src/",
@@ -94,7 +96,10 @@
     "generate-exports": "node scripts/generate-exports.mjs",
     "generate-packages": "node scripts/generate-packages.mjs",
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
-    "sync-readme": "node scripts/sync-readme.mjs"
+    "sync-readme": "node scripts/sync-readme.mjs",
+    "test": "yarn run test-types && yarn run build-craco",
+    "test-build": "yarn run build-craco",
+    "test-types": "tsc --noEmit"
   },
   "eslintConfig": {
     "extends": [
--- a/client/package.json
+++ b/client/package.json
@@ -86,8 +86,7 @@
   "scripts": {
     "setup": "git submodule update --init && yarn build-wasm && yarn generate",
     "start": "yarn generate && craco start",
-    "build": "yarn generate && yarn run build-craco",
-    "build-craco": "GENERATE_SOURCEMAP=false craco build",
+    "build": "yarn generate && GENERATE_SOURCEMAP=false craco build",
     "build-wasm": "build=\"../wasm/build.sh\" && chmod +x $build && $build --update",
     "format": "prettier --write src/",
     "check-format": "prettier --check src/",
--- a/client/package.json
+++ b/client/package.json
@@ -35,7 +35,7 @@
     "crypto-browserify": "^3.12.0",
     "jszip": "^3.10.1",
     "mocha": "^10.0.0",
-    "monaco-editor": "=0.36.1",
+    "monaco-editor": "=0.37.1",
     "pako": "^2.1.0",
     "prettier": "^2.7.1",
     "process": "^0.11.10",
--- a/client/package.json
+++ b/client/package.json
@@ -97,7 +97,6 @@
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
     "sync-readme": "node scripts/sync-readme.mjs",
     "test": "yarn test-types && yarn build",
-    "test-build": "yarn run build-craco",
     "test-types": "tsc --noEmit"
   },
   "eslintConfig": {
--- a/client/package.json
+++ b/client/package.json
@@ -96,7 +96,7 @@
     "generate-packages": "node scripts/generate-packages.mjs",
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
     "sync-readme": "node scripts/sync-readme.mjs",
-    "test": "yarn run test-types && yarn run build-craco",
+    "test": "yarn test-types && yarn build",
     "test-build": "yarn run build-craco",
     "test-types": "tsc --noEmit"
   },
--- a/client/src/utils/decorators/updatable.ts
+++ b/client/src/utils/decorators/updatable.ts
@@ -12,7 +12,7 @@ export type Updatable<T> = {
 
 /** Recursive `onDidChange${propertyName}` method types */
 export type OnDidChangePropertyRecursive<T, U = FlattenObject<T>> = {
-  [K in keyof U as `${typeof PROPS.ON_DID_CHANGE}${Capitalize<K>}`]: (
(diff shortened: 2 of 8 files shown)


rogaldh:
@acheroncrypto take a look, please

client/package.json
```diff
@@ -54,7 +54,7 @@
     "shiki": "^0.14.4",
     "stream-browserify": "^3.0.0",
     "styled-components": "^5.3.3",
-    "typescript": "=4.9.5",
+    "typescript": "=5.0.4",
```
acheroncrypto, line 57: Why this version specifically?

acheroncrypto (commented):
`typescript` is an actual dependency of the project (not a dev-dependency). Its version is also loosely tied to the `monaco-editor`'s version for consistent behavior. I remember trying to upgrade them a long time ago, and there were some problems with the newer versions. The `monaco-editor` version we use is the last version that uses TS v4:

https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/package.json#L38

https://github.com/microsoft/monaco-editor/blob/v0.36.1/package.json#L64

I think we should update them both (`monaco-editor` can be updated in a separate PR)'; otherwise the editor will use TS v4, but we'll transpile using TS v5. The transpilation is referring to the user-created clients:

https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/src/utils/client/client.ts#L1

https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/src/utils/client/client.ts#L99-L102

client/package.json
rogaldh: This is the minimal version that satisfies `@solana/codecs-data-structures`

rogaldh:
> `typescript` is an actual dependency of the project (not a dev-dependency). Its version is also loosely tied to the `monaco-editor`'s version for consistent behavior. I remember trying to upgrade them a long time ago, and there were some problems with the newer versions. The `monaco-editor` version we use is the last version that uses TS v4:
> 
> https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/package.json#L38
> 
> https://github.com/microsoft/monaco-editor/blob/v0.36.1/package.json#L64
> 
> I think we should update them both (`monaco-editor` can be updated in a separate PR)'; otherwise the editor will use TS v4, but we'll transpile using TS v5. The transpilation is referring to the user-created clients:
> 
> https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/src/utils/client/client.ts#L1
> 
> https://github.com/solana-playground/solana-playground/blob/85094d3f19ddade19477b02e8a7f169a54c75a49/client/src/utils/client/client.ts#L99-L102

That is fair. I'll update Monaco within the current PR. It looks feasible

client/package.json
```diff
@@ -77,15 +77,17 @@
...
   },
   "engines": {
     "node": ">=22.20.0"
   },
   "scripts": {
     "setup": "git submodule update --init && yarn build-wasm && yarn generate",
     "start": "yarn generate && craco start",
-    "build": "yarn generate && GENERATE_SOURCEMAP=false craco build",
+    "build": "yarn generate && yarn run build-craco",
+    "build-craco": "GENERATE_SOURCEMAP=false craco build",
```
acheroncrypto: The `generate` script takes about <1% of the total build time, but it's crucial to run in case there is a change. Why are we separating them?

```suggestion
    "build": "yarn generate && GENERATE_SOURCEMAP=false craco build",
```

client/package.json
```diff
@@ -94,7 +96,10 @@
     "generate-exports": "node scripts/generate-exports.mjs",
     "generate-packages": "node scripts/generate-packages.mjs",
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
-    "sync-readme": "node scripts/sync-readme.mjs"
+    "sync-readme": "node scripts/sync-readme.mjs",
+    "test": "yarn run test-types && yarn run build-craco",
```
acheroncrypto:
```suggestion
    "test": "yarn test-types && yarn build",
```

client/package.json
```diff
@@ -94,7 +96,10 @@
...
+    "test-build": "yarn run build-craco",
```
acheroncrypto: What's the point of this script?

```suggestion
```

client/src/utils/types.ts
```diff
@@ -117,6 +117,9 @@ export type SyncOrAsync<T = unknown> = T | Promise<T>;
 /** A `Promise` or a callback that returns a `Promise` */
 export type Promisable<T> = Getable<SyncOrAsync<T>>;
 
+/** String-only keys of `T` */
+export type StringKeys<T> = Extract<keyof T, string>;
```
acheroncrypto: Nit: While this is technically more correct, I'd prefer to name this `KeyOf` similar to the `ValueOf` helper:

https://github.com/solana-playground/solana-playground/blob/178e8702bc2c73157a8b41b0743c0c739d66cb29/client/src/utils/types.ts#L1-L2

We don't really care about non-string keys.

Tiny nit: Move this at the top of the file, above `ValueOf`, since they're related.

```suggestion
/** String-only keys of `T` */
export type KeyOf<T> = Extract<keyof T, string>;
```

client/src/utils/decorators/updatable.ts
```diff
@@ -346,7 +346,9 @@ type JoinCapitalized<T extends string[]> = T extends [
...
 /** Map the property values to a union of tuples */
 type PropertiesToUnionOfTuples<T, Acc extends string[] = []> = {
-  [K in keyof T]: T[K] extends object
-    ? [[...Acc, K], T[K]] | PropertiesToUnionOfTuples<T[K], [...Acc, K]>
-    : [[...Acc, K], T[K]];
+  [K in keyof T]: K extends string
+    ? T[K] extends object
+      ? [[...Acc, K], T[K]] | PropertiesToUnionOfTuples<T[K], [...Acc, K]>
+      : [[...Acc, K], T[K]]
+    : never;
```
acheroncrypto: Can we also use the helper here?

```suggestion
  [K in StringKeys<T>]: T[K] extends object
    ? [[...Acc, K], T[K]] | PropertiesToUnionOfTuples<T[K], [...Acc, K]>
    : [[...Acc, K], T[K]];
```

acheroncrypto (changes requested):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

client/package.json
rogaldh: I see `yarn generate` in the `setup` script. It looks to me like a prerequisite to build the client, but I've kept a separate build script.
But your variant works as well.

client/package.json
rogaldh: Resovled

client/src/utils/types.ts
rogaldh: Ok

client/src/utils/decorators/updatable.ts
rogaldh: I tried this, and it won't work. KeyOf does Extract under the hood, but keys should be built at once, afaik. After that, the ternary branch is applied, and it multiplies the number of iterations for the compiler.
But I've [replaced](https://github.com/solana-playground/solana-playground/pull/445/commits/584c5117e416c3b2fff167d290c4203aaefafdfe) `extends string` with `extends KeyOf<T>` to narrow possible variants (not the broad string type).

Please check it out.

client/package.json
acheroncrypto: It's not only a prerequisite but also a requirement for other scenarios such as adding/removing impls (e.g. [`commands`](https://github.com/solana-playground/solana-playground/tree/f76f84715fa6f34ccb94f64fd29a4eca8b96b471/client/src/commands), [`settings`](https://github.com/solana-playground/solana-playground/tree/f76f84715fa6f34ccb94f64fd29a4eca8b96b471/client/src/settings) [`tutorials`](https://github.com/solana-playground/solana-playground/tree/f76f84715fa6f34ccb94f64fd29a4eca8b96b471/client/src/commands)). This is also why we also run it with the `start` script, not only the `setup` script.

acheroncrypto (approved):
Perfect!

Issue #117, opened 2023-02-23
Feature idea: Pyth price account address helper


mcintyre94:
The test UI has really nice helpers for filling in addresses, and I think a Pyth one could be a good addition. Pyth requires passing in the address of a price account, and these are kinda fiddly to find. They're also network-dependent

I think we could have a Pyth option in the account dropdown (alongside eg Random, from seeds etc), and then present an autocomplete search where you can type eg "SOL/USD" and get the public key for SOL/USD on the currently selected network. I think this would make testing programs that use Pyth lots easier.

I'm not sure if there's an official version but we could use the seahorse CSV file for the data: https://github.com/ameliatastic/seahorse-lang/blob/main/data/pyth.csv

acheroncrypto:
This sounds great but also a lot of potential unused data, so ideally we would want to lazy load the Pyth data only when it's necessary. I can work on this after #113.

acheroncrypto:
Finally was able to add this in in https://github.com/solana-playground/solana-playground/commit/08db5a76ffcf9d5e711d7b7cc3e9f785b2eb615c

If you check the commit, you'll see how easy it was to add it but it's only because I've rewritten the whole test UI to make it extendible(took so much longer than I anticipated). There are still so many possible improvements but I think Pyth accounts helper is a great addition, thank you for the idea!

![test-ui-pyth](https://github.com/solana-playground/solana-playground/assets/98934430/2201a3be-2868-468b-b02c-cf9bc9edb8f2)

PR #446, opened 2026-05-02
test-ui: add cached token and mint account generators

### LINKED ISSUE #175: Add token and mint account generators in test UI
Similar to Pyth generators(https://github.com/solana-playground/solana-playground/issues/117), we should also support token and mint account helpers.

### Requirements

- Show the token and mint accounts that the user owns
- Implement some sort of caching to not refetch accounts each time

### Optional requirements

- Show recommendations based on source code, i.e. if a token account is specified with a `token::mint` constraint and the `mint` account is already populated, we should recommend the token accounts owned by the user for that specific mint at the top(or even autofill?).
- Show the top mint accounts based on the current cluster

bit2swaz:
## Problem

The Test UI already supports helper-driven public key inputs such as Pyth accounts, but it does not provide similar helpers for SPL token accounts or mint accounts.

This makes testing token-related instructions slower and more error-prone because users need to manually discover and paste the relevant addresses each time.

Closes #175  
Related to #117

## Solution

- add wallet-owned token account suggestions to the existing generator flow
- derive mint suggestions from the wallet's owned token accounts
- cache fetched results per cluster + wallet and reuse in-flight requests
- surface new `Token accounts` and `Mint accounts` groups in the existing Test UI SearchBar flow
- persist selected token and mint values through the normal instruction generator lifecycle
- add focused tests for normalization and cache behavior

## Testing

- [x] `npx tsx --test src/utils/program-interaction/generators/token.test.ts`
- [x] targeted Prettier check on touched files
- [x] editor diagnostics clean on touched source files

## Limitations

- mint suggestions are currently derived from the wallet's owned token accounts rather than scanning mint-authority-owned mints
- cluster-wide top mint recommendations are not included in this PR
- full client-wide formatting / type-checking could not be completed in this checkout because local client dependencies are not installed


bit2swaz:
@acheroncrypto could you review this?

scutuatua-crypto (approved):

scutuatua-crypto (approved):

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,174 @@
...
+            amount?: string;
+            uiAmountString?: string;
+          };
+        };
+      };
+    };
+  };
+}
+
+type OwnerLike = { toBase58(): string };
```
acheroncrypto: Could you use the existing types instead of re-defining everything from scratch?

client/src/utils/program-interaction/program-interaction.ts
```diff
@@ -110,6 +117,25 @@ export class PgProgramInteraction {
...
+        if (!owner) return [];
+
+        const accounts =
+          await PgConnection.current.getParsedTokenAccountsByOwner(owner, {
+            programId: TOKEN_PROGRAM_ID,
+          });
+        return accounts.value;
+      },
+    });
+  };
```
acheroncrypto: Seperating implementations like this is a good practice generally speaking, but I don't think it's needed here. It also makes things inconsistent since other generators don't do this. Could you refactor this and add 2 separate methods (`getTokenAccounts` and `getMintAccounts`)?


```suggestion
  /** {@link getTokenAccounts} */
  static getTokenAccounts = getTokenAccounts;

  /** {@link getMintAccounts} */
  static getMintAccounts = getMintAccounts;
```

client/src/views/sidebar/test/Component/Instruction/InstructionInput.tsx
```diff
@@ -307,6 +307,36 @@ const getSearchBarProps = (
       closeButton: true,
     });
 
+    if (PgWallet.current && !accountProps.isSigner) {
```
acheroncrypto, line 310: Showing token and mint accounts after "From seed" and before "All programs" makes sense. Could you also roughly keep this order in other places that you added the impl to? It doesn't have to be exact, but you can put them above "Pyth" for instance. Only commenting here to not spam the same comment.

acheroncrypto (changes requested):
Thank you for this PR! This is indeed a very useful feature.

Have you tested this locally in the UI and confirmed it works as expected? In any case, it looks good overall, but I do have some minor recommendations to make the feature fit in more to the current codebase.

bit2swaz:
thanks for the review. yes, i tested this locally in the UI and confirmed it works as expected. i ran the app locally, opened the spl token vault tutorial, built and deployed it on devnet, switched to the test sidebar, and verified that the relevant public-key inputs now show both `Token accounts` and `Mint accounts`.

refactored everything else as requested as well :)

client/src/utils/program-interaction/generators/token.test.ts
acheroncrypto, line 1: This file is not checked in CI and therefore could easily break. Just FYI, it's not required to add tests.

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
+import type { Connection, PublicKey } from "@solana/web3.js";
+
```
acheroncrypto: Nit: We want to avoid importing `@solana/web3.js` directly. Could you import `PgWeb3` instead?

```suggestion
import type { PgWeb3 } from "../../web3";
```

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+  timestamp?: number;
+  promise?: Promise<TokenAccountsData>;
+};
+
+const EMPTY_TOKEN_ACCOUNTS: TokenAccountsData = {
+  tokenAccounts: [],
+  mintAccounts: [],
+};
+
+const TOKEN_PROGRAM_ID = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
```
acheroncrypto: What about token-2022?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+
+const DEFAULT_CACHE_TTL = 30;
+
+const getNow = () => Math.floor(Date.now() / 1000);
```
acheroncrypto: Can use [`PgCommon.getUnixTimestamp`](https://github.com/solana-playground/solana-playground/blob/5869d3739eccb85c6607669f9ba7930c06d6dcba/client/src/utils/common.ts#L518) and inline this in `TokenAccountsCache.getOrInit`.

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+
+export class TokenAccountsCache {
+  private readonly _entries: Record<string, CacheEntry> = {};
+
+  constructor(private readonly _ttlSeconds = DEFAULT_CACHE_TTL) {}
+
+  get(key: string) {
+    return this._entries[key]?.data ?? EMPTY_TOKEN_ACCOUNTS;
+  }
+
```
acheroncrypto: Is this method used anywhere? If not, please remove.

```suggestion
```

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+  ) {
+    const entry = (this._entries[key] ??= {});
+
+    if (
+      entry.data &&
+      entry.timestamp !== undefined &&
+      now <= entry.timestamp + this._ttlSeconds
+    ) {
+      return entry.data;
+    }
```
acheroncrypto: Thoughts on invalidating the cache when the user makes a new transaction? You can listen for transaction send events with [`PgTx.onDidSend`](https://github.com/solana-playground/solana-playground/blob/5869d3739eccb85c6607669f9ba7930c06d6dcba/client/src/utils/tx.ts#L119).

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    if (entry.promise) return await entry.promise;
+
+    entry.promise = fetcher()
+      .then((data) => {
+        entry.data = data;
+        entry.timestamp = now;
+        entry.promise = undefined;
```
acheroncrypto: Nit: could you either set to `null` or `delete` the key instead?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+        return data;
+      })
+      .catch((error) => {
+        entry.promise = undefined;
```
acheroncrypto: Same here.

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+  clear(key?: string) {
+    if (key) {
+      delete this._entries[key];
+      return;
+    }
+
+    for (const existingKey of Object.keys(this._entries)) {
+      delete this._entries[existingKey];
+    }
+  }
```
acheroncrypto: Same as `get`, could you remove if it's not used anywhere?

```suggestion
```

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+const tokenAccountsCache = new TokenAccountsCache();
+
+export const normalizeTokenAccounts = (
+  accounts: ParsedTokenAccounts
+): TokenAccountsData => {
+  const tokenAccounts: TokenAccount[] = [];
+  const mintAccounts: string[] = [];
+  const seenMints = new Set<string>();
```
acheroncrypto: Why have 2 variables for the same data? Both seem to be set at the exact same place below:

```ts
if (seenMints.has(mint)) continue;
seenMints.add(mint);
mintAccounts.push(mint);
```

acheroncrypto (changes requested):
Thanks for addressing the previous issues! There are a couple more issues left. Consider adding [the new contribution guidelines](https://github.com/solana-playground/solana-playground/blob/5869d3739eccb85c6607669f9ba7930c06d6dcba/CONTRIBUTING.md#typescript) to your context. The TypeScript code in this repo is very opinionated in many aspects, so it's totally normal for you to not get it right the first time.

client/src/utils/program-interaction/generators/token.test.ts
bit2swaz: removed the file

client/src/utils/program-interaction/generators/index.ts
```diff
@@ -1,2 +1,3 @@
+export { getMintAccounts, getTokenAccounts } from "./token";
```
acheroncrypto: Could you move this below to preserve the alphabetical order?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
+import { PgCommon } from "../../common";
+import type { PgWeb3 } from "../../web3";
+
+/** A single SPL token account owned by the current wallet */
+export interface TokenAccount {
```
acheroncrypto: Does this type need to be exported? If not, let's make it private.
```suggestion
interface TokenAccount {
```

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+  address: string;
+  mint: string;
+  amount: string;
+}
+
+/** Token accounts and the unique mints derived from them */
+export interface TokenAccountsData {
```
acheroncrypto: Same here:
```suggestion
interface TokenAccountsData {
```

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+const clearTokenAccountsCache = () => {
+  for (const key in tokenAccountsCache) delete tokenAccountsCache[key];
+};
+
+/** Invalidate the cache when a transaction is sent (registers only once). */
+const setInvalidationListener = async () => {
+  if (isInvalidationListenerSet) return;
+  isInvalidationListenerSet = true;
+
+  const { PgTx } = await import("../../tx");
```
acheroncrypto: What's blocking us to import everything normally? `PgProgramInteraction` already depends on `PgTx`, so there is no benefit for lazy-loading.

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/** Fetch the current wallet's Token and Token-2022 accounts. */
+const fetchTokenAccounts = async (
+  owner: PgWeb3.PublicKey
+): Promise<ParsedTokenAccounts> => {
+  const [{ PgConnection }, { PgWeb3 }] = await Promise.all([
+    import("../../connection"),
+    import("../../web3"),
+  ]);
```
acheroncrypto: Same here. Both are already loaded at this time. Could you import normally?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+        programId: new PgWeb3.PublicKey(programId),
+      })
+    )
+  );
+
+  return results.flatMap((result) => result.value);
+};
+
+/** Derive sorted token accounts and unique mints from parsed accounts. */
+const normalizeTokenAccounts = (
```
acheroncrypto: Could you merge `fetchTokenAccounts` and `normalizeTokenAccounts` into a single function?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+  accounts: ParsedTokenAccounts
+): TokenAccountsData => {
```
acheroncrypto: Could you let TS infer the return types when possible?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  return { tokenAccounts, mintAccounts };
+};
+
+/** Get or fetch the wallet's token accounts, cached per cluster/owner. */
+const getOrInitTokenAccounts = async (): Promise<TokenAccountsData> => {
+  const [{ PgConnection }, { PgWallet }] = await Promise.all([
+    import("../../connection"),
+    import("../../wallet"),
+  ]);
```
acheroncrypto: These are also already loaded at this time. Could you import normally?

client/src/utils/program-interaction/generators/token.ts
```diff
@@ -0,0 +1,179 @@
...
+
+/** Get the current wallet's token accounts for the connected cluster. */
+export const getTokenAccounts = async () => {
+  return (await getOrInitTokenAccounts()).tokenAccounts;
+};
+
+/** Get the unique mints from the current wallet's token accounts. */
+export const getMintAccounts = async () => {
+  return (await getOrInitTokenAccounts()).mintAccounts;
+};
```
acheroncrypto: Could you move these functions above so that they are the first functions from the top of the file (below the types and constants)?

acheroncrypto (commented):
Awesome! This is almost ready to merge. Last nitpicks.

acheroncrypto (approved):
Thank you for the contribution and sticking with me till the end!

PR #447, opened 2026-05-04
feat(client): deploy the client as a static bundle to Vercel

rogaldh:
## Context

This PR originally introduced a Node/Hono BFF + `X-API-Key` to gate the build endpoints, with a planned Turnstile follow-up. That direction was rolled back — the API stays public, no API keys — and the team requested deploying the client as a static bundle to Vercel instead. The Rust server stays on GAE; the browser talks to it directly.

## Problem

The CRA bundle includes a wasm toolchain build step that Vercel's runner doesn't have, so a plain "connect Git to Vercel" flow fails on every deploy. There's also no built-in way to point a Vercel build at a non-Vercel backend without baking the URL in at build time, and the existing `client/src/settings/server/server.ts` hardcoded `api.solpg.io` in production regardless of `REACT_APP_SERVER_URL`.

## Why

The team's other projects deploy their client apps to Vercel. This PR brings the playground client in line with that pattern — one consistent deployment story across projects, instead of a special case here. The Vercel workflow is on-demand (`workflow_dispatch`-only), not automatic; it doesn't modify the existing default pipeline, and pushes to master continue to deploy the server to GAE unchanged.

## Solution

The build runs on a GitHub Actions runner (or local dev machine) that already has the wasm toolchain, then `vercel deploy --prebuilt` ships the prebuilt output. Vercel runs no build of its own — `client/vercel.json` keeps `buildCommand: null` and the project's "Ignored Build Step" short-circuits Vercel-side builds.

- **Manual GitHub Actions workflow** at `.github/workflows/_deploy_client_vercel.yml`, triggered via `workflow_dispatch`. Defaults to preview; promotion to prod alias is manual.
- **Local-deploy Makefile** at repo root — `vercel-bootstrap` (one-time link), `vercel-link-{production,preview}` (pull env), `deploy-client-to-vercel-{production,preview}` (build + ship). Only `VERCEL_TOKEN` needs to be in shell.
- **Endpoint routing** — `client/src/settings/server/server.ts` now respects `REACT_APP_SERVER_URL` at build time for non-share endpoints. Share endpoints (`/share/*`, `/new`) stay hardcoded to `api.solpg.io` so shared snippets remain discoverable from one store.
- **BFF rollback** — explicit `Revert "feat(bff)..."` commit. The `web/` Hono service, `X-API-Key` middleware, `prod_next` compose profile, and related artifacts are undone with the audit trail preserved.
- **Docs** at `docs/deploy-client-vercel.md` — one-time setup, deploy paths, endpoint routing, CORS prerequisite.

Defense against non-browser abuse (build cost, spam to `/new`) isn't added here — conscious trade-off, separate follow-up if it becomes a problem.


rogaldh:
@acheroncrypto please check it out

acheroncrypto (changes requested):
This looks like a grand effort, but first, I want to ask, why do we want this in the first place? We currently allow local usage, and this change will make the server unusable with other clients. You could perhaps solve this by adding a system to share the API keys, but that would complicate the implementation even further. Do you have any plans for the API keys in the future?

Also, there seems to be a bunch of changes that have nothing to do with adding API keys to the server (e.g. adding program hashes to build responses). Could you please make separate PRs for them?

PS: It would be great if you could give a quick heads-up before you work on big features like this one. It would suck to reject a PR that someone spent a lot of time and effort on. Both messaging me directly and creating a GitHub issue would work.

rogaldh:
> This looks like a grand effort, but first, I want to ask, why do we want this in the first place? We currently allow local usage, and this change will make the server unusable with other clients. You could perhaps solve this by adding a system to share the API keys, but that would complicate the implementation even further. Do you have any plans for the API keys in the future?
> 
> Also, there seems to be a bunch of changes that have nothing to do with adding API keys to the server (e.g. adding program hashes to build responses). Could you please make separate PRs for them?
> 
> PS: It would be great if you could give a quick heads-up before you work on big features like this one. It would suck to reject a PR that someone spent a lot of time and effort on. Both messaging me directly and creating a GitHub issue would work.

The main goal was to introduce API_KEY to allow protecting usage in production.
- API_KEY is configured - protection works
- API_KEY is not - it works as well (to enable local usage)

I agree that a bunch of other changes came to make the whole thing work. I'll extract improvements into separate PRs to make this feature as small as possible.

acheroncrypto (commented):
> The main goal was to introduce API_KEY to allow protecting usage in production.

Well, I see that's what you want to do, but I'm asking why you want to do that. Why do you want to "protect usage in production", or in other words, block other clients from working with the public server?

rogaldh:
> > The main goal was to introduce API_KEY to allow protecting usage in production.
> 
> Well, I see that's what you want to do, but I'm asking why you want to do that. Why do you want to "protect usage in production", or in other words, block other clients from working with the public server?

Yes. I look at it from the API perspective. Having no key allows anybody to call the endpoint and basically deny the service. Especially with the limited number of slots to perform builds.

acheroncrypto:
> > > The main goal was to introduce API_KEY to allow protecting usage in production.
> > 
> > 
> > Well, I see that's what you want to do, but I'm asking why you want to do that. Why do you want to "protect usage in production", or in other words, block other clients from working with the public server?
> 
> Yes. I look at it from the API perspective. Having no key allows anybody to call the endpoint and basically deny the service. Especially with the limited number of slots to perform builds.

Yeah, I get the concern, but it seems to me that this PR does not have user-based authentication, meaning it doesn't even solve the "deny the service" problem you mentioned, since people can just spam the API from the allowed client. Feel free to correct me if I'm wrong about that (haven't reviewed in detail yet).

Similarly, having a public API with no authentication does not necessarily mean that people can easily spam the service. The `/build` route should be rate-limited regardless of authentication.

Additionally, pretty much every other nonprofit playground, such as Rust Playground and Go Playground, also allows public access. Note that when I say "public access/API", I mean access without any kind of authorization headers and not referring to the public/private IP kind of distinction (which would be a bit silly in this context). For instance, you can just:

```sh
curl -s -X POST https://play.rust-lang.org/execute \
  -H "Content-Type: application/json" \
  -d '{
    "channel":"stable",
    "mode":"debug",
    "edition":"2024",
    "crateType":"bin",
    "tests":false,
    "code":"fn main() { println!(\"Hello, world!\"); }"
  }'
```

and receive:

```
{"success":true,"exitDetail":"Exited with status 0","stdout":"Hello, world!\n","stderr":"   Compiling playground v0.0.1 (/playground)\n    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.66s\n     Running `target/debug/playground`\n"}
```

This is a feature, not a bug. Therefore, what do you say to continuing to allow the same supported clients (including beta.solpg.io, local/VSCode extension)? I don't think there is any real benefit for people to DoS the build server; they'd just be wasting their time/money.

However, if it ever becomes a real issue, which is a big if on a better hardware (the current prod server is quite low spec due to lack of funding), then we can consider the options to fix it.

If my assessment is correct (the PR not actually resolving the mentioned problem), fixing this "properly" would probably require us to integrate third-party authorization systems (e.g. GitHub), which imo would damage the UX massively.

client/src/settings/server/server.ts
```diff
@@ -12,11 +12,10 @@ export const server = [
     ],
     default:
       process.env.NODE_ENV === "production"
-        ? "https://api.solpg.io"
-        : // Docker builds use this environment variable to set the server URL
-          // to the production API (instead of local) if the user has not yet
-          // built the server image
-          process.env.REACT_APP_SERVER_URL ?? "http://localhost:8080",
+        ? process.env.REACT_APP_SAME_ORIGIN === "1"
+          ? "." // relative URL — requests go to the same-origin BFF
```
acheroncrypto: This is infinitely better than all the previous empty string workarounds. Just wanted to note this, but not reviewing fully until our discussion resolves.

rogaldh:
Quick context update — the initial description was thin on motivation, so let me lay out the thinking and why this PR is changing shape.

The original BFF + `API_KEY` idea was meant to cover a gap that CORS alone leaves open: non-browser callers. The server-side `CLIENT_URLS` allowlist only works because browsers honour the `Origin` header; anything that doesn't (curl, scripts, server-side code) just sends the request and the server processes it. An `API_KEY` held inside a BFF would have closed that gap, because a key is something you have to possess, not a label you can self-declare like `Origin`. That distinction is the thing — they're not interchangeable; they cover different threats.

That direction was ruled out: the API is intentionally public, no API keys. I want to be explicit that this is a *policy* decision, not a "CORS already covers it" one.

The next step I had in mind was a Turnstile (or similar) integration to make automated spam from the official client harder. The BFF was the natural place to mount that, since the server-side process can verify the token before forwarding upstream.

Then the migration to Vercel was requested, which dissolves what's left of the original design. With the client served as a static bundle, there's no place for an `API_KEY` anyway — a key in a public bundle is readable in JS, same as no key — so in this state, none is introduced. The captcha mount point goes away with the BFF too. `CLIENT_URLS` keeps its existing role, with the Vercel deployment origin added to the list as an endorsed caller.

So this PR is being transformed into the flow that deploys the client as a static bundle to Vercel. The backend stays public. One thing worth flagging: defence against non-browser abuse (build cost, spam to `/new`) isn't in place under the new architecture — a conscious trade-off rather than an oversight, and worth a follow-up if it becomes a problem.

.github/workflows/_deploy_client_vercel.yml
acheroncrypto, line 1: Could you rename this file to `deploy-client.yml`? The reason is, if we decide to host it on another platform, there shouldn't be a need to change the file name. The `vercel` suffix only makes sense if there are multiple platforms to deploy the client to. The `_` prefix is also redundant.

client/src/settings/server/server.ts
```diff
@@ -2,21 +2,24 @@ import { GITHUB_URL } from "../../constants";
...
 
+const customServerUrl = process.env.REACT_APP_SERVER_URL;
+
 export const server = [
   createSetting({
     id: "server.endpoint",
     description: "Build server URL",
     values: [
+      ...(customServerUrl
+        ? [{ name: "Solana Playground", value: customServerUrl }]
```
acheroncrypto: The foundation-hosted server should not use a custom URL. It should be hardcoded so that all other clients, including local and beta.solpg.io, also use it. It should also always be the default on prod builds, since it will be faster.

client/src/settings/server/server.ts
```diff
@@ -2,21 +2,24 @@ import { GITHUB_URL } from "../../constants";
...
+        : []),
+      { name: "Solpg Playground API", value: "https://api.solpg.io" },
```
acheroncrypto: "Solpg" is already an abbrevation of "Solana Playground", so the word "Playground" shouldn't be used after it.

It's also redundant to append "API" to a setting called "Server Endpoint", since all of them are APIs.

I think it would be better to mention who hosts the server, given all can be considered Solana Playground anyway:

- Local
- Solana Foundation
- SolPg

acheroncrypto: Could you keep "Local" one at the top? All settings are roughly ordered from smallest to biggest. For example:

- `connection.endpoint`:

    https://github.com/solana-playground/solana-playground/blob/e5478fbbb0f8de9407da71e2ab764b1434e6fab2/client/src/settings/connection/connection.ts#L16-L22

    (based on importance in terms of prod/dev diff)

- `connection.commitment`

    https://github.com/solana-playground/solana-playground/blob/e5478fbbb0f8de9407da71e2ab764b1434e6fab2/client/src/settings/connection/connection.ts#L9

    (based on confirmations):

    `processed`: 0
    `confirmed`: 1
    `finalized`: 32

Others are usually ordered alphabetically, which can also be considered a form of smallest to biggest.

client/docs/deploy-client-vercel.md
acheroncrypto, line 1: Could you move this inside the `client` folder? Both a separate file and including it to the existing [README](https://github.com/solana-playground/solana-playground/blob/e5478fbbb0f8de9407da71e2ab764b1434e6fab2/client/README.md) file would work.

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,47 @@
...
+Each target has a matching `vercel-link-{production,preview}` that's run automatically as a prerequisite.
+
+`REACT_APP_SERVER_URL` must be configured in Vercel for whichever environment you're deploying to (Production for `-production`, Preview for `-preview`, or both).
+
+## Endpoint routing
+
+Split is deliberate:
+
+- **Build/deploy is configurable** (`REACT_APP_SERVER_URL` at build time, also user-overridable via the `server.endpoint` setting) — so self-hosted forks can route builds to their own backend.
+- **Share (`/share/*`, `/new`) is hardcoded to `https://api.solpg.io`** — so shared snippets stay discoverable from one canonical store regardless of where the client is hosted.
```
acheroncrypto: Could you update this to say something like all routes except shares are configurable? Because there will most likely be new endpoints soon™, meaning it will no longer be just build/deploy.

acheroncrypto (changes requested):
> Then the migration to Vercel was requested, which dissolves what's left of the original design. With the client served as a static bundle, there's no place for an `API_KEY` anyway — a key in a public bundle is readable in JS, same as no key — so in this state, none is introduced. The captcha mount point goes away with the BFF too. `CLIENT_URLS` keeps its existing role, with the Vercel deployment origin added to the list as an endorsed caller.

That wasn't my request but nice! The current client (beta.solpg.io) also just hosts static files. Simpler is better here.

> So this PR is being transformed into the flow that deploys the client as a static bundle to Vercel. The backend stays public. One thing worth flagging: defence against non-browser abuse (build cost, spam to `/new`) isn't in place under the new architecture — a conscious trade-off rather than an oversight, and worth a follow-up if it becomes a problem.

As I mentioned in https://github.com/solana-playground/solana-playground/pull/447#issuecomment-4459666520, prod server endpoints should be rate-limited (especially `/build`).

.github/workflows/_deploy_client_vercel.yml
rogaldh: Prefix means that this workflow is a reusable one. To be included within another one.

client/src/settings/server/server.ts
rogaldh: Do you mean to hardcode the address within the repo?

.github/workflows/_deploy_client_vercel.yml
acheroncrypto: It seems like the workflow can be used standalone, and there is no other workflow that reuses it.

The repo also prefers kebab-case file names over snake_case ones when possible.

client/src/settings/server/server.ts
acheroncrypto: Yes

client/scripts/vercel-install.sh
```diff
@@ -0,0 +1,34 @@
...
+  echo ">>> Build cache MISS: no cached cargo, installing rustup from scratch"
+fi
+
+if ! command -v cargo >/dev/null 2>&1; then
+  # Pin matches wasm/rust-toolchain.toml so the initial `cargo install wasm-pack`
+  # has a working toolchain before per-package rust-toolchain.toml files apply.
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+    | sh -s -- -y --default-toolchain 1.75.0 --profile minimal --target wasm32-unknown-unknown
+fi
+rustup default 1.75.0
```
acheroncrypto: Could you read [the `rust-toolchain.toml` file](https://github.com/solana-playground/solana-playground/blob/a4e72c7993ec50c5eca78d5d4c182d118bd0d63d/wasm/rust-toolchain.toml) to get the version instead of hardcoding?

client/src/settings/server/server.ts
```diff
@@ -2,21 +2,24 @@ import { GITHUB_URL } from "../../constants";
 import { PgCommon } from "../../utils";
 import { createSetting } from "../create";
 
+const solanaFoundationServerUrl =
+  process.env.REACT_APP_SOLANA_FOUNDATION_SERVER_URL;
```
acheroncrypto: Why is this still an environment variable?

client/README.md
```diff
@@ -16,6 +16,10 @@ Start the dev server:
 yarn start
 ```
 
+## Deployment
+
+For Vercel deployment instructions, see [Deploy client to Vercel](docs/deploy-client-vercel.md).
+
```
acheroncrypto: Nit: Move this to the bottom, since it's the least useful for most people.

client/src/settings/server/server.ts
rogaldh: Why don't you like the variant with the env variable, but the one we should provide? The address might change in the future, and hardcoding it requires a commit to alter it. That makes it harder to maintain.

.github/workflows/_deploy_client_vercel.yml
rogaldh: I managed to configure integration with Vercel, and it requires no extra workflow. I'll close that thread

client/src/settings/server/server.ts
rogaldh: Backend is not deployed to GAE yet. I think that having this env variable benefits the project. At the moment, the backend would be available; the only need is to specify this env at the configuration and redeploy.
I do not see flaws with this approach

client/src/settings/server/server.ts
acheroncrypto: Because it doesn't make sense to change the address of a public API. This is like changing api.devnet.solana.com; sure, it's possible, but one would need a strong reason to break everyone that uses it, so it's better to avoid it.

Longer reasoning: there are 2 possible scenarios:

1. URL never changes

    If this is the case, then it makes sense the hardcode it.

2. URL changes

    If the URL changes, you will break all the people who use it, including the people who use the playground client, because this is a setting that is persisted. We'd need a commit to fix this breakage and migrate everyone to the new URL. This migration code will need to live in the codebase for a long time, possibly forever. Having a hardcoded URL would be much simpler. Therefore, even in this scenario, it makes more sense to hardcode it.

I think you could just make it something similar to these:

- api.devnet.solana.com
- api.testnet.solana.com

and never have to change it. Something like:

- api.playground...
- api.play...

client/src/settings/server/server.ts
acheroncrypto: Answered in https://github.com/solana-playground/solana-playground/pull/447#discussion_r3431687903.

> I do not see flaws with this approach

On top of the flaws I mentioned above, another flaw is that other clients will need to pass this hardcoded value anyway. The URL should never change after being public. This is not some private internal URL that we can change whenever we want; it's a public server.

client/src/settings/server/server.ts
rogaldh: What about the local development when you want to use a local service instead of the public one?

rogaldh: > Because it doesn't make sense to change the address of a public API.

Just keep it within the envs and do not change it. Having an env gives you flexibility.

client/src/settings/server/server.ts
rogaldh: Usually this is done via having a `.env.example` with a proper value to use

client/README.md
rogaldh: Fixed

client/scripts/vercel-install.sh
rogaldh: Done

client/src/settings/server/server.ts
acheroncrypto: > What about the local development when you want to use a local service instead of the public one?

During local development, you automatically have the the "Local" server setting set by default. You can also just change the setting to "Local" if you previously had a different value.

With the `.env` approach, the user will need to know the foundation-hosted endpoint address to be able to use the public endpoint locally, since it doesn't exist in the codebase.

Local Docker (without server) will also not work by default since the foundation-hosted endpoint is not set.

> > Because it doesn't make sense to change the address of a public API.
> 
> Just keep it within the envs and do not change it. Having an env gives you flexibility.

Did you read the reasoning in the previous comment? It doesn't give you flexibility. You *will* break people if you change the address.

rogaldh: > > What about the local development when you want to use a local service instead of the public one?
> 
> During local development, you automatically have the the "Local" server setting set by default. You can also just change the setting to "Local" if you previously had a different value.
> 
> With the `.env` approach, the user will need to know the foundation-hosted endpoint address to be able to use the public endpoint locally, since it doesn't exist in the codebase.
> 
> Local Docker (without server) will also not work by default since the foundation-hosted endpoint is not set.
> 
> > > Because it doesn't make sense to change the address of a public API.
> > 
> > 
> > Just keep it within the envs and do not change it. Having an env gives you flexibility.
> 
> Did you read the reasoning in the previous comment? It doesn't give you flexibility. You _will_ break people if you change the address.

We use GAE and want to test the feature in a production mode (when the client is built for production).
Two reasons:
- Domain is not bind to the server yet, we use a temporary one at the moment.
- I want to build and test production version of the client. Criterion NODE_ENV=production does not allow to launch automated tests with a local server.

acheroncrypto: > * Domain is not bind to the server yet, we use a temporary one at the moment.

Then we can wait until adding "Solana Foundation" as a setting because, as I mentioned multiple times already, changing the URL will not make the existing users switch to the new one.

Adding a new setting value is also not required for you to test it.

> * I want to build and test production version of the client. Criterion NODE_ENV=production does not allow to launch automated tests with a local server.

Why not just set the endpoint in tests then? It's configurable:

```ts
PgSettings.server.endpoint = process.env....
```

or if you're adding UI tests:

```
setting set server.endpoint <URL>
```

rogaldh: Regarding the tests, I mean the automatic tests for the client, not the tests within the client.

acheroncrypto: And you can't use any of the solutions mentioned above?

rogaldh: It requires adding code that changes the configuration within the running client. I'd prefer to avoid it.

acheroncrypto: So instead of adding code to the tests, you think it's better to add code to the actual source to make tests work? If so, I don't follow this argument. The whole point of settings is that they are configurable. Tests should follow the source code, not the other way around.

rogaldh: > Tests should follow the source code, not the other way around

We are talking about just one environment variable, which allows you to configure the endpoint for a remote server that should receive build requests. That is a configuration improvement, and it eases up automated tests for this functionality, especially when we are testing the app locally.

Yes, I'm proposing this as it does cost nothing, as it is up to you what host to put into the variable (or not to put it at all).

If I were in your shoes, I'd agree with this statement
>So instead of adding code to the tests, you think it's better to add code to the actual source to make tests work? 

if the change would be bigger than adding: `const host = process.env.REMOTE_HOST ?? 'falback'`.

rogaldh:
@acheroncrypto I've hardcoded the address for the build server.

client/src/settings/server/server.ts
```diff
@@ -2,17 +2,21 @@ import { GITHUB_URL } from "../../constants";
 import { PgCommon } from "../../utils";
 import { createSetting } from "../create";
 
+const SOLANA_FOUNDATION_SERVER_URL =
+  "https://playground-server-dot-analytics-324114.de.r.appspot.com";
```
acheroncrypto: > @acheroncrypto I've hardcoded the address for the build server.

Is this URL stable? Meaning, will it stay the same as long as we want? And is it unique? Meaning, is it impossible for other people to claim the same appspot.com URL if you release it (no longer use it)?

If the answer to all of the questions above is "yes", then we can merge this as is. If not, to merge this ASAP, I'm willing to accept your environment variable solution **until** we have the actual foundation domain. Deal?

Another option would be to hold off adding the "Solana Foundation" value until we have the actual foundation URL (as mentioned in https://github.com/solana-playground/solana-playground/pull/447#discussion_r3464400067), but this is probably the least ideal solution for the testing case.

rogaldh: `analytics-324114` is generated based on the project ID in GCP. Is unique per project (globally).
`de` is generated based on the project's region during project creation. Could not be changed.

The only part that could be changed is `playground-server-`. But it is taken from app.yaml.

acheroncrypto: Thanks for the info. It looks pretty ugly though. It wouldn't be the end of the world to merge as is, but I think it's probably better to have your environment variable approach until we have the proper URL. Or we could also revert all changes to this file until that happens, but I'm guessing you don't want this?

rogaldh: Yeah. Just want to test things out on Vercel's Preview build.
I could add this. Names are just an example, will use proper ones.
```
const URL = process.env.FNDN_URL ?? 'https://api.playground.solana.com'
```
This helps to use the Appspot one currently and the final one later. Are you ok with that?

rogaldh: Added here. Please take a look: https://github.com/solana-playground/solana-playground/pull/447/commits/19351f470a182161581ce1e7d2b8e4b5681b29b9

client/src/settings/server/server.ts
```diff
@@ -2,17 +2,24 @@ import { GITHUB_URL } from "../../constants";
 import { PgCommon } from "../../utils";
 import { createSetting } from "../create";
 
+const SOLANA_FOUNDATION_SERVER_URL =
+  process.env.REACT_APP_FOUNDATION_SERVER_URL ??
```
acheroncrypto: Do you see a problem with using the existing `REACT_APP_SERVER_URL` for all cases (instead of `REACT_APP_FOUNDATION_SERVER_URL`)?

client/src/settings/server/server.ts
acheroncrypto: In line with the previous comment, if you're fine with that, can we keep the `api.solpg.io` default until we have the proper URL? You can override the default with an environment variable like you wanted from the start. Something like:

```ts
default:
  // The default value can be overridden with the `REACT_APP_SERVER_URL`
  // environment variable.
  //
  // If the value does not exists in the `values` section, the setting will
  // be shown as a "Custom" value.
  //
  // Docker builds also use this environment variable to set the server URL
  // to the production API (instead of local) if the user has not yet built
  // the server image.
  process.env.REACT_APP_SERVER_URL ??
  (process.env.NODE_ENV === "production"
    ? "https://api.solpg.io"
    : "http://localhost:8080"),
```

client/Dockerfile
```diff
@@ -21,7 +21,7 @@ ONBUILD COPY --from=solana-playground-server /home/${USER}/.cargo /root/.cargo
 FROM ${IMAGE} AS server-false
 
 # Set build server URL to the production API
-ONBUILD ENV REACT_APP_SERVER_URL="https://api.solpg.io"
+ONBUILD ENV REACT_APP_SERVER_URL="https://playground-server-dot-analytics-324114.de.r.appspot.com"
```
acheroncrypto: Could you revert this change until we have the real domain?
```suggestion
ONBUILD ENV REACT_APP_SERVER_URL="https://api.solpg.io"
```

client/src/settings/server/server.ts
rogaldh: Just to clarify. The original task was to allow playground users to build programs on the server deployed by Solana Foundation. Currently, the instance is live on *.[appspot.com](http://appspot.com/) url. The solution that you've proposed allows me to test it but does not allow users to use the new server.
I see two ways forward:

1. Push it as is on the *.[appspot.com](http://appspot.com/) URL so users can start using it, and merge the PR.
2. Merge the PR with the solution proposed by you, test it on staging, and hold off until you find agreement with SF on the URL

How would you like to proceed?

acheroncrypto: > The original task was to allow playground users to build programs on the server deployed by Solana Foundation.

Yes, that's what I wanted from the start as well, but I have doubts about whether the test domain will always work. It would be nice to avoid unnecessary migration code afterwards.

Additionally, although not as important, I was expecting the new server to be much faster than the current one. Based on what you've said in https://github.com/solana-playground/solana-playground/pull/463, this doesn't seem to be the case, at least currently.

> I see two ways forward:
> 
> 1. Push it as is on the *.[appspot.com](http://appspot.com/) URL so users can start using it, and merge the PR.
> 2. Merge the PR with the solution proposed by you, test it on staging, and hold off until you find agreement with SF on the URL
> 
> How would you like to proceed?

Option 2 sounds to be the better option as of now.

rogaldh: > Additionally, although not as important, I was expecting the new server to be much faster than the current one. Based on what you've said in https://github.com/solana-playground/solana-playground/pull/463, this doesn't seem to be the case, at least currently.

Faster in deploying the server or building the program? Anyway, that is a topic for optimisation and is not related to the current PR.

I added a commit that follows Option 2

client/Dockerfile
rogaldh: done

client/src/settings/server/server.ts
rogaldh: Fixed, but the comment was shortened

client/src/settings/server/server.ts
acheroncrypto: > Faster in deploying the server or building the program?

Program build time is, of course, the more important part, but note that the performance of deploying the server and building a program are more or less affected by the same things: mostly CPU.

> Anyway, that is a topic for optimisation and is not related to the current PR.

I wouldn't really consider this topic related to optimization; it's the comparison of the servers themselves. As I mentioned earlier, the current server, `api.solpg.io`, had been downgraded to a very low spec server about 2 years ago (due to lack of funding). This is why I was expecting a much better performance with the new deployment(s).

I think it may be worth exploring different configurations that could make Rust builds faster (which is a big chunk of what the deployment process spends time on), such as faster CPU and storage. More cores could help right now, but soon, we'll be limiting core usage of individual program builds, likely by hardcoding it to a specific value, meaning having more cores is unlikely to help.

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,43 @@
+# Deploy the client to Vercel
+
+Vercel's native Git integration auto-deploys: `master` → production; any other branch → preview. Makefile targets exist as a local escape hatch.
```
acheroncrypto, line 3: Not a blocker to merge, but auto-deploying client while deploying the server manually (via tags) could make things out-of-sync.

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,43 @@
...
+
+`installCommand` = `bash scripts/vercel-install.sh` (rustup + `wasm/build.sh` + `yarn install`); `buildCommand` = `yarn build`. Wasm must precede `yarn install` because `client/package.json` has `file://../wasm/*/pkg` deps that don't exist until `wasm-pack` runs.
+
+## Verified Vercel project settings
+
+| Setting | Value |
+| --- | --- |
+| Plan / Build Machine | Enterprise + **Enhanced** (~18 min cold) |
```
acheroncrypto: This can get outdated pretty quickly:

```suggestion
| Plan / Build Machine | Enterprise + **Enhanced** |
```

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,43 @@
...
+3. Production Branch: `master`. Ignored Build Step: Automatic.
+4. Node.js Version: `22.x`.
+5. Account Settings → Tokens: team-scoped token, `export VERCEL_TOKEN=...` locally for the Makefile targets.
+6. Link the local checkout (from repo root):
+
+   ```sh
+   VERCEL_PROJECT_ID=prj_xxx make vercel-bootstrap
+   ```
+
+Add the Vercel deployment origin to the GAE server's `client_urls` or CORS will reject every request.
```
acheroncrypto:
```suggestion
Add the Vercel deployment origin to the server's [`PG_CLIENT_URLS`](https://github.com/solana-playground/solana-playground/blob/cd5555155c61572c8c49fb351890519af9e493ef/.env.example#L3) environment variable or CORS will reject every request.
```

client/scripts/vercel-install.sh
```diff
@@ -0,0 +1,37 @@
...
+  # Version comes from wasm/rust-toolchain.toml so the initial `cargo install wasm-pack`
+  # has a working toolchain before per-package rust-toolchain.toml files apply.
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+    | sh -s -- -y --default-toolchain "$RUST_VERSION" --profile minimal --target wasm32-unknown-unknown
+fi
+rustup default "$RUST_VERSION"
+
+bash ../wasm/build.sh
+
+yarn install --frozen-lockfile
```
acheroncrypto, line 37: We're not using [the `setup` script](https://github.com/solana-playground/solana-playground/blob/cd5555155c61572c8c49fb351890519af9e493ef/client/package.json#L88) here because of the custom Rust caching logic inside `node_modules`? It feels we would be able to simplify this script quite a bit if there was a way to tell Vercel what to cache.

(this comment is fine to ignore if you don't want to do it)

client/vercel.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "framework": null,
+  "installCommand": "bash scripts/vercel-install.sh",
```
acheroncrypto, line 4: Does Vercel handle submodules automatically? I don't see submodules being initialized anywhere.

https://github.com/solana-playground/solana-playground/blob/cd5555155c61572c8c49fb351890519af9e493ef/client/package.json#L88

client/Makefile.vercel
acheroncrypto, line 1: Could you please move this to the `client` dir? The root `Makefile` can stay.

acheroncrypto (commented):
Looking good for the most part. I left a couple more comments, mostly nits. The only critically important one to address is the one related to submodules.

client/docs/deploy-client-vercel.md
rogaldh: I agree that it is better to have these processes in sync. This behaviour was requested by Jonas and is partially based on the approach we discussed earlier regarding the state of the master.

client/docs/deploy-client-vercel.md
rogaldh: Will fix

client/Makefile.vercel
rogaldh: Ok

client/vercel.json
rogaldh: Yes, Vercel does that automatically as long as the submodule is accessible via HTTPS: https://vercel.com/docs/builds/build-features#git-submodules

client/docs/deploy-client-vercel.md
rogaldh: fixed

client/scripts/vercel-install.sh
rogaldh: Let's keep the script as is

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,43 @@
...
+## Verified Vercel project settings
+
+| Setting | Value |
+| --- | --- |
+| Plan / Build Machine | Enterprise + **Enhanced** |
+| Framework Preset | Other |
+| Root Directory | `client` |
+| Production Branch | `master` |
+| Ignored Build Step | Automatic |
+| Node | `22.x` (project); `^22.20.0` (`package.json` engines) |
```
acheroncrypto, line 16: I believe Vercel uses `engines.node` automatically. This can also get outdated pretty easily. It's better to have the `package.json` as the only source of truth.

```suggestion
```

client/docs/deploy-client-vercel.md
```diff
@@ -0,0 +1,43 @@
...
+
+## One-time setup
+
+1. Create the project. Framework: Other. Root Directory: `client`.
+2. Build Machine: Enhanced on Enterprise; default on Pro.
+3. Production Branch: `master`. Ignored Build Step: Automatic.
+4. Node.js Version: `22.x`.
```
acheroncrypto, line 23: Same here:

```suggestion
```

client/package.json
```diff
@@ -96,8 +97,9 @@
     "generate-packages": "node scripts/generate-packages.mjs",
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
     "sync-readme": "node scripts/sync-readme.mjs",
-    "test": "yarn test-types && yarn build",
-    "test-types": "tsc --noEmit"
+    "test": "yarn test-types && yarn test-unit && yarn build",
+    "test-types": "tsc --noEmit",
+    "test-unit": "craco test --watchAll=false"
```
acheroncrypto, line 102: This can break very easily, as it is not checked in CI.

acheroncrypto (approved):
I left a couple more comments, but since this has already gone through a decent amount of revisions, I will merge it as is. Feel free to address them later.

To reiterate (https://github.com/solana-playground/solana-playground/pull/447#discussion_r3654138061) just in case, the deployments must stay in sync. Having the same logic for both would be a lot easier to manage. We should resolve this issue before switching to the real domain.

PR #449, opened 2026-05-06
Warn during deployments before reusing stale binary after a failed build

rogaldh:
## Surface "deploy after failed build" client-side instead of mutating server state

image

### Background

After a successful build, `programs/<uuid>/solpg.so` sits on disk on the build server. If the user then submits a build that **fails to compile**, `program::build` early-returns without touching the previous artifact, and `GET /deploy/:uuid` returns the previous successful binary — the playground silently deploys stale code.

## Why not fix this server-side

An earlier version of this branch deleted `solpg.so` before each `cargo-build-sbf` invocation. After review feedback from @acheroncrypto the direction changed, and the server-side work has been dropped from this PR:

- "Leave the previous artifact alone on compile failure" is what `cargo`, `solana-program`, and `anchor` do locally. A user running `anchor build` and hitting a compile error keeps the prior `target/deploy/*.so` on disk.
- The playground is gaining more frameworks and versions. Each ships its own build CLI with its own assumptions about artifact lifecycle. The less the server overrides those, the fewer surprises across frameworks.
- The user-facing problem isn't the file on disk — it's the **silent deploy**. The user doesn't know they're about to upload an old binary. That's a UX concern, not a build-server concern.

This PR's diff against `master` is therefore purely client-side; the server tree is untouched.

## What this PR does

1. **Tracks the last build error on the client.** `PgProgramInfo` gains a persisted `lastBuildError: string | null` field, carried in both `ProgramInfo` (in-memory) and `SerializedProgramInfo` (storage). The build command sets it to `result.stderr` when stderr matches the server's own compile-failure check — `stderr.includes("error: could not compile")`, mirroring `server/src/program.rs` — and clears it back to `null` on the next successful build. The check is intentionally not gated on `result.uuid`: the server only returns a `uuid` on the **first** build per client session and `null` thereafter, regardless of success, so it can't double as a success signal. The flag is persisted because the server's cached `solpg.so` survives page reloads — the warning signal has to survive with it.

2. **Warns and asks for confirmation on deploy.** In `checkProgram` (the deploy command's pre-check), if `lastBuildError` is set and there's no user-imported `.so` to use instead, the terminal prints `Warning: Your last build failed. Deploying now will upload the last successful build, not the current sources.`, then prompts `Deploy the previously built binary anyway? [yes/no]` with `default: "no"`. Declining throws `Deployment cancelled: last build failed.` so the existing `try/finally` resets deploy state cleanly.

3. **Ensures the new prompt is visible.** `waitForUserInput` in `terminal.ts` now calls `scrollToBottom()` after `focus()`. `focus()` alone only scrolls to the cursor; if the user had scrolled up while reading the failed-build output, the new prompt would land off-screen.

## Behavior matrix

| State | Behavior on deploy |
|---|---|
| Last build succeeded (or no prior failure recorded) | Deploys, no prompt. |
| Last build failed, no imported `.so` | Warning + `[yes/no]` confirmation, `default: "no"`. Decline ⇒ deploy aborts. Confirm ⇒ deploys the server-cached binary — same end result as before this PR, but now opt-in. |
| Last build failed, imported `.so` present | Deploys the imported binary, no prompt — the server-cached binary isn't read. |

### DIFF
--- a/client/src/commands/deploy/deploy.ts
+++ b/client/src/commands/deploy/deploy.ts
@@ -70,6 +70,24 @@ async function checkProgram() {
     await PgCommand.build.execute();
   }
 
+  // The server keeps the previous `solpg.so` after a failed compile, so /deploy
+  // would silently upload the stale binary. Make that explicit and let the user
+  // opt in. Only relevant when we'd actually use the server-cached binary.
+  if (
+    PgProgramInfo.lastBuildError &&
+    !PgProgramInfo.importedProgram?.buffer.length
+  ) {
+    PgTerminal.println(
+      "Warning: Your last build failed. Deploying now will upload the last successful build, not the current sources."
+    );
+    const term = await PgTerminal.get();
+    const proceed = await term.waitForInput(
+      "Deploy the previously built binary anyway?",
+      { confirm: true, default: "no" }
+    );
+    if (!proceed) throw new Error("Deployment cancelled: last build failed.");
+  }
+
   if (!PgProgramInfo.pk) {
     throw new Error(
       "Program ID not found. Go to 'Build & Deploy' tab and set the program ID."
--- a/client/src/commands/deploy/deploy.ts
+++ b/client/src/commands/deploy/deploy.ts
@@ -70,11 +70,11 @@ async function checkProgram() {
     await PgCommand.build.execute();
   }
 
-  // The server keeps the previous `solpg.so` after a failed compile, so /deploy
-  // would silently upload the stale binary. Make that explicit and let the user
-  // opt in. Only relevant when we'd actually use the server-cached binary.
+  // The server keeps the previous program binary after a failed compilation, so
+  // /deploy would silently return the stale binary. Ask the user about whether
+  // to proceed with the deployment using the server-cached binary.
   if (
-    PgProgramInfo.lastBuildError &&
+    PgProgramInfo.lastBuildFailed &&
     !PgProgramInfo.importedProgram?.buffer.length
   ) {
     PgTerminal.println(
--- a/client/src/utils/program-info.ts
+++ b/client/src/utils/program-info.ts
@@ -31,6 +31,8 @@ type ProgramInfo = Nullable<{
     buffer: Buffer;
     fileName: string;
   };
+  /** stderr from the last failed build; cleared on success. Persisted so a reload doesn't drop the warning. */
+  lastBuildError: string;
 }>;
 
 /** Serialized program info that's used in storage */
@@ -39,6 +41,7 @@ type SerializedProgramInfo = Nullable<{
   kp: Array<number>;
   customPk: string;
   idl: Idl;
+  lastBuildError: string;
 }>;
 
 const defaultState: ProgramInfo = {
@@ -47,6 +50,7 @@ const defaultState: ProgramInfo = {
   customPk: null,
   idl: null,
   importedProgram: null,
+  lastBuildError: null,
 };
 
 const storage = {
@@ -88,6 +92,7 @@ const storage = {
       idl: state.idl,
       kp: state.kp ? Array.from(state.kp.secretKey) : null,
       customPk: state.customPk?.toBase58() ?? null,
+      lastBuildError: state.lastBuildError,
     };
 
     await PgExplorer.fs.writeFile(this.PATH, JSON.stringify(serializedState));
--- a/client/src/utils/program-info.ts
+++ b/client/src/utils/program-info.ts
@@ -31,7 +31,7 @@ type ProgramInfo = Nullable<{
     buffer: Buffer;
     fileName: string;
   };
-  /** Whether the last build failed; cleared on success. Persisted so a reload doesn't drop the warning. */
+  /** Whether the last build failed */
   lastBuildFailed: boolean;
 }>;
 
--- a/client/src/utils/program-info.ts
+++ b/client/src/utils/program-info.ts
@@ -31,8 +31,8 @@ type ProgramInfo = Nullable<{
     buffer: Buffer;
     fileName: string;
   };
-  /** stderr from the last failed build; cleared on success. Persisted so a reload doesn't drop the warning. */
-  lastBuildError: string;
+  /** Whether the last build failed; cleared on success. Persisted so a reload doesn't drop the warning. */
+  lastBuildFailed: boolean;
 }>;
 
 /** Serialized program info that's used in storage */
@@ -41,7 +41,7 @@ type SerializedProgramInfo = Nullable<{
   kp: Array<number>;
   customPk: string;
   idl: Idl;
-  lastBuildError: string;
+  lastBuildFailed: boolean;
 }>;
 
 const defaultState: ProgramInfo = {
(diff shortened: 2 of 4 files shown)


rogaldh:
@acheroncrypto please, check it out

acheroncrypto (commented):
Yea, it can be considered a problem, but it's also how Rust, Solana, and Anchor behave. Considering we'll support more frameworks and versions soon™, I think it would be beneficial for us to keep the default tooling behavior (the behavior you'd have during local development), especially server-side.

What do you think of showing a warning client-side if trying to deploy after a compile error instead? We can ask for a confirmation to proceed with the deployment for instance.

Also, thank you for the concise PR description(s)! They are appreciated.

rogaldh:
> Yea, it can be considered a problem, but it's also how Rust, Solana, and Anchor behave. Considering we'll support more frameworks and versions soon™, I think it would be beneficial for us to keep the default tooling behavior (the behavior you'd have during local development), especially server-side.
> 
> What do you think of showing a warning client-side if trying to deploy after a compile error instead? We can ask for a confirmation to proceed with the deployment for instance.
> 
> Also, thank you for the concise PR description(s)! They are appreciated.

Gotcha. Thanks for context. I agree with that approach, pretty neat idea.

I played with it a bit, do you like this one? @acheroncrypto 
image


Please let me know your thoughts. I'll update this PR accordingly.

acheroncrypto:
> I played with it a bit, do you like this one? @acheroncrypto

Yep, that looks pretty good. Just a tiny nit: I've recently been using "Warning: ..." and "*Error: ..." for warnings and errors instead of highlighting the whole text in order to avoid showing too much color (the keywords get automatic highlighting). Just FYI, also fine as is.

Also, unrelated to this PR, I've noticed the airdrop on devnet has failed for some reason. This has been a massive pain point. Can we fix this (without the user having to go to another website just to get an airdrop)? Maybe the API should be more generous on requests coming from the foundation-hosted client for example. "Internal error" is also not a great response.

rogaldh:
> Also, unrelated to this PR, I've noticed the airdrop on devnet has failed for some reason. This has been a massive pain point. Can we fix this (without the user having to go to another website just to get an airdrop)? Maybe the API should be more generous on requests coming from the foundation-hosted client for example. "Internal error" is also not a great response.

Definitely a different thing. I'll try to check it out.

client/src/commands/deploy/deploy.ts
```diff
@@ -70,6 +70,24 @@ async function checkProgram() {
     await PgCommand.build.execute();
   }
 
+  // The server keeps the previous `solpg.so` after a failed compile, so /deploy
+  // would silently upload the stale binary. Make that explicit and let the user
+  // opt in. Only relevant when we'd actually use the server-cached binary.
```
acheroncrypto: Even if it's a comment, it would be great to not have comments about server internals in client (`solpg.so` is not exposed), so better to refer it as the "program binary". Also made some adjustments to make the intent clearer:

```suggestion
  // The server keeps the previous program binary after a failed compilation, so
  // /deploy would silently return the stale binary. Ask the user about whether
  // to proceed with the deployment using the server-cached binary.
```

client/src/utils/terminal/terminal.ts
```diff
@@ -331,6 +331,11 @@ export class PgTerm {
     this.scrollToCursor();
   }
 
+  /** Focus only; `scrollToCursor()`'s formula desyncs the DOM scroll from xterm's state and leaves the next render at the wrong position. */
+  focusOnly() {
+    this._xterm.focus();
+  }
```
acheroncrypto: I've noticed the wrong position behavior in some cases too, but I'm not sure if the issue is with the `scrollToCursor` impl or with the way we call it in `waitForInput` (the latter seems more likely). In any case, I think it would be better to inline this in `waitForInput` and add a comment explaining why we don't use `focus` directly.

client/src/utils/program-info.ts
```diff
@@ -31,6 +31,8 @@ type ProgramInfo = Nullable<{
     buffer: Buffer;
     fileName: string;
   };
+  /** stderr from the last failed build; cleared on success. Persisted so a reload doesn't drop the warning. */
+  lastBuildError: string;
```
acheroncrypto: We don't seem to be using the string value anywhere, so it seems redundant to have it and also save it to storage (output can be long). Why not store a simple boolean instead?

acheroncrypto (commented):
Nice! Just a few minor issues, and we can get this in after.

client/src/utils/terminal/terminal.ts
rogaldh: Do you mean to rewrite the comment?

acheroncrypto: No, remove `focusOnly` and inline `this._xterm.focus()` in `waitForInput` with a comment explaining the reason for not calling `focus` (of `PgTerm` which has `scrollToCursor`).

rogaldh: ok

client/src/utils/terminal/terminal.ts
```diff
@@ -451,7 +451,9 @@ export class PgTerm {
         : number
       : string
   > {
-    this.focus();
+    // Avoid `this.focus()` here: its `scrollToCursor()` desyncs the DOM scroll from xterm's state and leaves the next render at the wrong position.
```
acheroncrypto: Nit: This line is 148 chars long. Could you keep comments at around 80 chars per line max?

Also, I don't think the problem with `scrollToCursor` *always* happens, so let's encode this knowledge with "may".

```suggestion
    // Avoid `this.focus()` here because it calls `scrollToCursor`, which may
    // desync the DOM scroll from `xterm`'s state and leave the next render at
    // the wrong position.
```

client/src/utils/program-info.ts
```diff
@@ -31,6 +31,8 @@ type ProgramInfo = Nullable<{
     buffer: Buffer;
     fileName: string;
   };
+  /** Whether the last build failed; cleared on success. Persisted so a reload doesn't drop the warning. */
```
acheroncrypto: Nit: This comment is also very much above the 80 char limit (107).

I think storing this kind of build info in `PgProgramInfo` makes sense. There is no need to create another type just to store this build state (maybe also `PgGlobal.buildLoading`). However, I have a few comments regarding this choice:

> cleared on success

`PgProgramInfo` itself doesn't manage this behavior, so let's not mention it here.

> Persisted so a reload doesn't drop the warning.

Similar to above, `PgProgramInfo` itself doesn't know about this "warning". Other fields don't mention whether they are persisted or not, so let's remove this too.

```suggestion
  /** Whether the last build failed */
```

I think the original comment would have made sense if the data were stored only for the `build` command, but since `PgProgramInfo` usage is not limited to the `build` command, doc comments should take that into account.

rogaldh: fixed

client/src/utils/terminal/terminal.ts
rogaldh: fixed

client/src/commands/deploy/deploy.ts
rogaldh: fixed

acheroncrypto (approved):
Looks great, thank you!

Just FYI: when the changes are trivial (e.g. comment changes), I'd usually commit the suggestions to merge ASAP, but I can't do it because you're sending PRs from an organization fork. Nothing wrong with that, but I wanted to let you know just in case.

PR #450, opened 2026-05-06
fix(server): improve slot handling

rogaldh:
### Problem

The `concurrency_id` slot could leak when `/build` failed abnormally - a panic in the blocking task, or the request future being dropped (client disconnect, cancellation). The imperative cleanup after `.await` never ran, the slot stayed marked taken, and enough leaks exhausted the pool with "Failed to find concurrency id" until restart.

### Solution

BuildState::reserve() returns a SlotGuard whose Drop releases the slot — runs on every exit path.

- Slot table switched to `std::sync::Mutex` so Drop can release it.
- `spawn_blocking(...).await.expect(...)` → ? mapping JoinError to anyhow!.
- Mutex poison handled symmetrically via `unwrap_or_else(|e| e.into_inner())`.

Tests cover normal-drop, panic, cancellation, unique ids per reservation, and that an over-capacity reservation blocks until a slot frees.

### DIFF
--- a/server/src/program.rs
+++ b/server/src/program.rs
@@ -113,7 +113,11 @@ pub fn build(
 
     // Check compile errors
     let stderr = String::from_utf8(output.stderr)?;
-    if stderr.rfind("error: could not compile").is_some() {
+    let rust_compile_failed = stderr.rfind("error: could not compile").is_some();
+    // `cargo-build-sbf` post-link checks (e.g. stack-size verifier) emit lines starting with
+    // "Error:" but exit 0, so an exit-code or rustc-only check would miss them.
+    let sbf_link_failed = stderr.lines().any(|line| line.starts_with("Error: "));
+    if rust_compile_failed || sbf_link_failed {
         return Ok((stderr, None));
     }
 
--- a/server/src/routes/build.rs
+++ b/server/src/routes/build.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use anchor_syn::idl::types::Idl;
 use anyhow::anyhow;
@@ -8,7 +8,7 @@ use axum::{
 };
 use serde::{Deserialize, Serialize};
 use tokio::{
-    sync::{Mutex, Semaphore, SemaphorePermit},
+    sync::{Semaphore, SemaphorePermit},
     task,
 };
 use uuid::Uuid;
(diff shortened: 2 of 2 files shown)


rogaldh:
@acheroncrypto, please check it out

scutuatua-crypto (approved):

server/src/routes/build.rs
```diff
@@ -72,34 +72,64 @@ impl BuildState {
...
+    async fn reserve(&self) -> Result<SlotGuard<'_>> {
+        let permit = self
+            .sem
+            .acquire()
+            .await
+            .map_err(|e| anyhow!("Failed to acquire `Semaphore`: {e}"))?;
+        let mut ids = self.ids.lock().unwrap_or_else(|e| e.into_inner());
+        let id = ids
+            .iter()
+            .position(|used| !used)
```
acheroncrypto, line 86: This definitely looks cleaner than what I was doing ([`.enumerate().find_map()`](https://github.com/solana-playground/solana-playground/blob/f40dc269d50443d5371fb6d0e09f75764f79879a/server/src/routes/build.rs#L156-L157)).

server/src/routes/build.rs
```diff
@@ -72,34 +72,64 @@ impl BuildState {
...
+
+impl SlotGuard<'_> {
+    fn id(&self) -> usize {
+        self.id
+    }
+}
+
+impl Drop for SlotGuard<'_> {
+    fn drop(&mut self) {
+        let mut ids = self.ids.lock().unwrap_or_else(|e| e.into_inner());
```
acheroncrypto, line 113: How would the `Mutex` get poisoned? I wasn't sure whether to do ignore the poison or not, but eventually sided with not ignoring it:

https://github.com/solana-playground/solana-playground/blob/f40dc269d50443d5371fb6d0e09f75764f79879a/server/src/routes/build.rs#L173-L177

Something about recovering potentially corrupt data feels wrong. Having said that, I don't have a strong position on this matter, but unless you have a scenario where the poison happens, we can always choose to ignore it later on (if it ever happens). We'd quickly learn about it since the server would essentially stall (at least the builds).

server/src/program.rs
```diff
@@ -113,7 +113,11 @@ pub fn build(
 
     // Check compile errors
     let stderr = String::from_utf8(output.stderr)?;
-    if stderr.rfind("error: could not compile").is_some() {
+    let rust_compile_failed = stderr.rfind("error: could not compile").is_some();
+    // `cargo-build-sbf` post-link checks (e.g. stack-size verifier) emit lines starting with
+    // "Error:" but exit 0, so an exit-code or rustc-only check would miss them.
+    let sbf_link_failed = stderr.lines().any(|line| line.starts_with("Error: "));
+    if rust_compile_failed || sbf_link_failed {
```
acheroncrypto, line 120: The reason why it exits with status code 0 is because those are not *actually* errors; they're just logs that the compiler considers to be *potential* errors. If you want to learn more about this, I explained the situation in more detail in:

- https://www.anchor-lang.com/docs/updates/release-notes/0-31-0#stack-memory-improvements
- https://github.com/solana-foundation/anchor/issues/3060#issuecomment-2197714568

Therefore, this change is incorrect because it makes the server return an error even if there is none (technically not returning an error but functionally so).
```suggestion
    if stderr.rfind("error: could not compile").is_some() {
```

acheroncrypto (changes requested):
Thank you for the swift fix, but this feature was a work in progress, which is why I hadn't messaged you about it yet. I usually push minimal versions first and then iterate on it later on, so if you see a recent change, it would be better to give it a couple days or message me before making changes to the same place to avoid unnecessary work and conflicts.

The problem should be fixed by https://github.com/solana-playground/solana-playground/commit/171d0d0b9f931a3186878beb38efbd9e2ad17ad3, which is slightly different than this PR. One difference (`Iter::position`) is clearly better handled in this PR, but I'm not sure about the poison issue. I left a comment about them both.

On the other hand, treating *potential* error logs as hard compile errors is not right. I also left a comment describing that issue.

PR #377, opened 2025-07-16
feat: postgres migration

jacobcreech:

PR #451, opened 2026-05-08
feat(db): finalise the migration from mongo to postgres

rogaldh:
## Problem

PR #377 (postgres migration) had been open since mid-2025, fell behind master, and would lose all mongo-era share URLs on cutover — old `_id` (24-char hex ObjectId) lookups don't fit a postgres `UUID` column.

## Solution

- Rebased: `postgres_migration` onto current master.
- Feedback: resolve comments from #377.
- Dual-format share IDs: new shares use `Uuid::new_v4()`; old mongo IDs are preserved via a new `legacy_id TEXT` column with a partial `UNIQUE` index. `GET /share/:id` parses UUIDs first, falls back to a `legacy_id` lookup for 24-char hex inputs. Server-side helper isolated in `db/id.rs`.
- Client matches the contract: `PgShare.isValidId` accepts UUID or 24-char hex (mongo ObjectId), implemented via `uuid.validate` + a strict length+hex check (intentionally stricter than `PgCommon.isHex` to reject `0x`-prefixed and wrong-length inputs). Added `@types/jest` and a `test:unit` script.
- Tests: Suite for `is_legacy_id`.

> **Note:** this PR only adds the *capability* to preserve old IDs. Backfilling existing mongo share documents into postgres is out of scope and must be done as a separate one-shot data migration during cutover, with both DBs reachable.


acheroncrypto:
Do you have anything in mind that would make use of a database in the future? Asking because I'm considering getting rid of the DB completely as it causes needless confusion and doesn't do much.

rogaldh:
> Do you have anything in mind that would make use of a database in the future? Asking because I'm considering getting rid of the DB completely as it causes needless confusion and doesn't do much.

Except for the "share" functionality, nothing comes to mind.

rogaldh:
@acheroncrypto I've updated the PR according to the latest upstream changes. Please take a look

acheroncrypto (commented):
Thanks for continuing the work! The PR looks great for the most part, but I wonder whether the reason why you want to migrate to Postgres is because you want to run DB separately on GCP? Honestly, you don't even need to run the DB right now. As long as api.solpg.io runs, it can continue to be the only source of truth (the current plan is to run it at least until the end of this year).

The reason why I suggested for the foundation to run the server is because of how computationally expensive the builds are, not because of the shares or the DB, for that matter. Everything else is pretty light and likewise cheap to run.

In any case, this is just an open question: what would you think of getting rid of DB completely in favor of using a Git repo for the shares? Open to suggestions if you have alternatives. Btw, I'm not opposed to this change, just trying to gauge the best long-term solution.

rogaldh:
> Thanks for continuing the work! The PR looks great for the most part, but I wonder whether the reason why you want to migrate to Postgres is because you want to run DB separately on GCP? Honestly, you don't even need to run the DB right now. As long as api.solpg.io runs, it can continue to be the only source of truth (the current plan is to run it at least until the end of this year).
> 
> The reason why I suggested for the foundation to run the server is because of how computationally expensive the builds are, not because of the shares or the DB, for that matter. Everything else is pretty light and likewise cheap to run.
> 
> In any case, this is just an open question: what would you think of getting rid of DB completely in favor of using a Git repo for the shares? Open to suggestions if you have alternatives. Btw, I'm not opposed to this change, just trying to gauge the best long-term solution.

There is CloudSQL configured for Postgres already. That is why initially Jacob brought that. Switching to Mongo will delay the deployment.

Having no DB right now requires tweaking the project here and there, as sharing won't work without a DB. The intent was to use the original version of the project and not rely on a fork. That means allowing deployment without DB right now requires introducing changes that will affect your state, which I do not want to do.

If you are leaning towards removing the DB and sharing functionality, that is okay, but the best way for me is:
- To deploy the project in the current state with DB support
- Make changes within the separate PR to remove sharing and other related things

That would help migrate faster.

> As long as api.solpg.io runs, it can continue to be the only source of truth (the current plan is to run it at least until the end of this year).

I considered that. I think we can not just deploy api.solpg.io. Currently, anyone could request the server programmatically, which is undesirable. That is why I created another PR with the BFF layer to allow protecting it with an API_KEY.

acheroncrypto:
> I think we can not just deploy api.solpg.io. Currently, anyone could request the server programmatically, which is undesirable. 

Could you please explain why it's undesirable exactly? This is supposed to be a public server.

rogaldh:
> > I think we can not just deploy api.solpg.io. Currently, anyone could request the server programmatically, which is undesirable.
> 
> Could you please explain why it's undesirable exactly? This is supposed to be a public server.

Public, but that does not mean that it should not have authentication, right?
Build is costly, no way to prevent access to it might lead to denial of service.

acheroncrypto:
Since we're discussing the same issue on the other PR, I'll just link my last comment (https://github.com/solana-playground/solana-playground/pull/447#issuecomment-4459666520).

rogaldh:
> Since we're discussing the same issue on the other PR, I'll just link my last comment ([#447 (comment)](https://github.com/solana-playground/solana-playground/pull/447#issuecomment-4459666520)).

Okay. Let's continue there about this topic

What do you think about the DB migration itself?

acheroncrypto:
> What do you think about the DB migration itself?

My answer depends on whether the servers are going to be public (no authorization). If they're public, and if you're fine with using api.solpg.io for shares, I think the simplest solution would be to to continue to use that (no changes needed since [the client already handles this](https://github.com/solana-playground/solana-playground/blob/e84ea20f51fa4aa9efbeb5f8542d48a10cab704c/client/src/utils/server.ts#L130-L132)).

In the future, we may get rid of the DB altogether and use a Git repo for shares instead. Note that in https://github.com/solana-playground/solana-playground/pull/451#pullrequestreview-4287297636, I did **not** mention getting rid of the sharing functionality, I asked "what would you think of getting rid of DB completely in favor of using a Git repo for the shares?" In other words, find a way to get rid of DB while keeping the sharing functionality. I proposed using a Git repo for the shares, but that's up for debate.

If you're not fine with using api.solpg.io for shares, then I'd prefer for you to just run MongoDB instead of going for a managed SQL solution. We really don't need a managed DB for this, as the data complexity and workload are both extremely light. You can run this on pretty much any machine (even the same machine as the build server would work). Just need to do periodic backups, and we're good to go.

However, if

- the server is going to require client authorization
- the server won't work with beta.solpg.io
- you don't want to run MongoDB


then we kind of have to get this PR in. This would be my least preferred option though. I think the first option is by far the simplest since it doesn't require any changes and would work with all clients.

rogaldh:
> you don't want to run MongoDB

We talked about that with Jonas. They already pay for Postgres in the cloud, and having Mongo will require paying extra. We agreed on trying to launch the project with Postgres.

> If they're public, and if you're fine with using api.solpg.io for shares

Do I get this right you suggest using api.solpg.io for sharing, but for building, to use a different way to resolve address?

rogaldh:
Regarding this line: https://github.com/solana-playground/solana-playground/blob/e84ea20f51fa4aa9efbeb5f8542d48a10cab704c/client/src/utils/server.ts#L130-L132
Could I move that address into an env variable to make it configurable? Having the exact address within the code as a magic string is not that good. At least making it configurable might improve the server utility.

rogaldh:
>My answer depends on whether the servers are going to be public (no authorization). If they're public, and if you're fine with using api.solpg.io for shares, I think the simplest solution would be to to continue to use that (no changes needed since [the client already handles this](https://github.com/solana-playground/solana-playground/blob/e84ea20f51fa4aa9efbeb5f8542d48a10cab704c/client/src/utils/server.ts#L130-L132)).

We discussed that, and at this point, to launch it ASAP, the server will be public, yes. But I do not like the intermediary state, when the client app is launched on Solana's domain, and the share functionality relies on api.solpg.io.

At first, don't get me wrong, I am trying to sync origin and fork as much as possible. Solana is committing to deploying Playground from the fork on their own infrastructure - that's already agreed. So there shouldn't be a huge drift in functionality.
As I've said before, there is already a managed Postgres there and switching to Mongo creates a drag and also should be paid for. I proposed to use Mongo, but that proposal was reverted.
At the moment of launching the app(client) on Solana's domain, we need to use the share functionality that runs on the Solana's infrastructure.
Even if we try to have an intermediary state, when the client uses api.solpg.io, we get stuck with several problems:
- configuration would be harder (we need to remember that sharing is somewhere else)
- testing becomes harder (we have to remember that our main scenario is not the one with client and server on the same domain)
- server and client have functionality which we do not use (we have to cover unhappy states for them via tests or patching the code)
- share is not just a simple api. Unlike the RPCs or data apis, sharing means that users store their data on the server and having this data on an external server brings legal consequences. Also, it is worth mentioning that even if we use api.solpg.io for some time and someone shares data there, we'll need to transfer it after.

acheroncrypto:
> > you don't want to run MongoDB
> 
> We talked about that with Jonas. They already pay for Postgres in the cloud, and having Mongo will require paying extra. We agreed on trying to launch the project with Postgres.

I've already addressed this in my previous reply (https://github.com/solana-playground/solana-playground/pull/451#issuecomment-4467643746). Please re-read this part:

> If you're not fine with using api.solpg.io for shares, then I'd prefer for you to just run MongoDB instead of going for a managed SQL solution. We really don't need a managed DB for this, as the data complexity and workload are both extremely light. You can run this on pretty much any machine (even the same machine as the build server would work). Just need to do periodic backups, and we're good to go.

---

> > If they're public, and if you're fine with using api.solpg.io for shares
> 
> Do I get this right you suggest using api.solpg.io for sharing, but for building, to use a different way to resolve address?

Mostly yes, and this is already implemented. As I mentioned on Discord, all you have to do is to make changes to the `server.endpoint` setting, and everything will work on all clients.

And to clarify the logic: use api.solpg.io for endpoints that use the DB, and use the configured server for everything else, not just build (I do have plans to add other non-DB endpoints soon™).

> Regarding this line:
> 
> https://github.com/solana-playground/solana-playground/blob/e84ea20f51fa4aa9efbeb5f8542d48a10cab704c/client/src/utils/server.ts#L130-L132
> 
> 
> Could I move that address into an env variable to make it configurable? Having the exact address within the code as a magic string is not that good. At least making it configurable might improve the server utility.

The whole reason why it's hardcoded is because the intended behavior is to make every client use the same DB server so that everyone has access to the same shares. Making that configurable would be the exact opposite of what we want.

> We discussed that, and at this point, to launch it ASAP, the server will be public, yes. But I do not like the intermediary state, when the client app is launched on Solana's domain, and the share functionality relies on api.solpg.io.

Consider this: Rust Playground uses GitHub gists for its share functionality. Does that mean Rust Playground is in an intermediary state? It's just one less thing for the foundation to worry about.

> As I've said before, there is already a managed Postgres there and switching to Mongo creates a drag and also should be paid for.

As I've already mentioned above, it doesn't need to be a drag and also doesn't need to be paid for.

> At the moment of launching the app(client) on Solana's domain, we need to use the share functionality that runs on the Solana's infrastructure.

Why is this a constraint? We could use any other service to store the shares in the future.

> * configuration would be harder (we need to remember that sharing is somewhere else)

I'm failing to understand how configuration would be harder. Not having to manage DB would make it much easier imo, and everything to support this usage is already implemented. Literally 0 configuration, pretty much no changes, and no need to worry about DB at all. How is this harder than managing a DB and also making a bunch of changes to make the integration work?

> * server and client have functionality which we do not use (we have to cover unhappy states for them via tests or patching the code)

Which client functionality do we not use? And why do we *have to* make changes?

> * share is not just a simple api. Unlike the RPCs or data apis, sharing means that users store their data on the server and having this data on an external server brings legal consequences.

What legal consequences? We do not store any personal data.

rogaldh:
> And to clarify the logic: use api.solpg.io for endpoints that use the DB, and use the configured server for everything else, not just build (I do have plans to add other non-DB endpoints soon™).

I get the logic. But my point is not about which endpoint routes where. It is about data sovereignty. Foundation deploys the service to its domain and is responsible for what runs there. That includes the data layer. A service that we host but cannot back up, restore, or audit is not something we can ship as an official Foundation service.
Also, it is nice to have it configurable.

> The whole reason why it's hardcoded is because the intended behavior is to make every client use the same DB server so that everyone has access to the same shares. Making that configurable would be the exact opposite of what we want.

That makes sense for the upstream public instance, and I'm not asking to change it there. But the same hardcoding makes it impossible to deploy the project as a separate service.

> Consider this: Rust Playground uses GitHub gists for its share functionality. Does that mean Rust Playground is in an intermediary state? It's just one less thing for the foundation to worry about.

By intermediary state, I refer to the situation when Playground is launched on Foundation's servers, and it stores data for shares at the api.solpg.io, the server which Foundation does not control.

> As I've already mentioned above, it doesn't need to be a drag and also doesn't need to be paid for.

Cost is also part of it. Managed Mongo on GCP would be a separate provider on top of what we already run — a new DB, new billing. Foundation already operates Postgres and has tooling.

> I'm failing to understand how configuration would be harder. Not having to manage DB would make it much easier imo, and everything to support this usage is already implemented. Literally 0 configuration, pretty much no changes, and no need to worry about DB at all. How is this harder than managing a DB and also making a bunch of changes to make the integration work?

You presume that you gonna implement the sharing functionality with no DB. But my logic is based on the situations where there is not yet a sharing functionality via Gists. I'm requested to migrate the Playground as a whole to the Google Platform (Application Engine + Vercel to be clear). I see the project with Mongo which Foundation does not use, but uses Postgres. And for me, a compromise to transfer backend there and keep sharing on the api.solpg.io is the intermediary state.

> Which client functionality do we not use? And why do we have to make changes?

Backend deployed to the GAE has endpoints to create and read shares and they become unusable with no DB on the foundation's site

> What legal consequences? We do not store any personal data.

Share that is created by the user - his data. Migration without preserving user-created shares is bad UX.
Also, I've already mentioned several times that Foundation wants to deploy the service to its domain. Foundation hosts under its domain. The server that you control is a third-party server/provider. Your server could be hacked, and data could be altered.
In this case, the Foundation is responsible for all the harm and undesired UX.

rogaldh:
I think we agreed on the compromise that DB not gonna be changed, and we aim to replace it with sharing via Gists. Could you, please, expand your idea?
I'll be able to help you build it. What do you think?

acheroncrypto:
> I get the logic. But my point is not about which endpoint routes where. It is about data sovereignty. Foundation deploys the service to its domain and is responsible for what runs there. That includes the data layer. A service that we host but cannot back up, restore, or audit is not something we can ship as an official Foundation service.
> Also, it is nice to have it configurable.

I've already given multiple examples about the hypocrisy (for the lack of a better word) of this statement on Discord. I just think you're being overly pedantic about this on a dev-only playground, especially when there are much more important security concerns elsewhere that could affect real users with real funds on mainnet. Even in the worst-case scenario, that being a hacked malicious share server, the damage in the real world would be so low that it could effectively be considered non-existent.

Again, to clarify my position, I wouldn't mention any of this if you actually had a strict policy of not using third-party tools. It just seems silly when the foundation-hosted services, **including solana.com**, have **already** been using the exact same shares that you've been concerned about the possibility of getting hacked. They've been doing this for many years.

That said, I think we see things very differently, so I don't see any point dragging this conversation longer.

> > As I've already mentioned above, it doesn't need to be a drag and also doesn't need to be paid for.
> 
> Cost is also part of it. Managed Mongo on GCP would be a separate provider on top of what we already run — a new DB, new billing. Foundation already operates Postgres and has tooling.

I mentioned that you don't have to run a managed DB for this simple service in both of my last 2 comments. Do you know how much hosting the shares has cost in total in the last 4 years? The answer is basically nothing (0). There is no need to scale something if it doesn't need to be scaled.

> > I'm failing to understand how configuration would be harder. Not having to manage DB would make it much easier imo, and everything to support this usage is already implemented. Literally 0 configuration, pretty much no changes, and no need to worry about DB at all. How is this harder than managing a DB and also making a bunch of changes to make the integration work?
> 
> You presume that you gonna implement the sharing functionality with no DB.

I didn't presume that. I presumed you'd use the code client as is (with the only change being the new server endpoint addition), which would be much simpler to manage for foundation than all other options.

> > What legal consequences? We do not store any personal data.
> 
> Share that is created by the user - his data.

No, it's not "his data"; the data has no owner, no personal information, and is willingly made public by the user. There is not even a way for a user to modify or delete this data because the data simply doesn't have an owner.

You said "having this data on an external server brings legal consequences". Such a definite statement that includes "brings legal consequences" should not be taken lightly. Therefore, I asked about the mentioned legal consequences, emphasizing this is non-personal public data. I specifically asked for it because there is a big difference between personal and non-personal data when it comes to law.

PR #452, opened 2026-05-08
fix(client): add noImplicitReturns to the config

rogaldh:
PR adds `noImplicitReturns` into tsconfig to be in sync with the config for VSCode and has better rules in general.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -97,7 +97,7 @@
     "generate-tutorials": "node scripts/generate-tutorials.mjs",
     "sync-readme": "node scripts/sync-readme.mjs",
     "test": "yarn test-types && yarn build",
-    "test-types": "tsc --noEmit"
+    "test-types": "yarn generate && tsc --noEmit"
   },
   "eslintConfig": {
     "extends": [
--- a/client/src/components/Editor/Home/Home.tsx
+++ b/client/src/components/Editor/Home/Home.tsx
@@ -197,7 +197,7 @@ const getSrc = (url: string) => {
   if (url.includes("youtube.com")) src = "youtube.png";
   else if (url.includes("dev.to")) src = "devto.png";
 
-  if (src) return "/icons/platforms/" + src;
+  return src ? "/icons/platforms/" + src : undefined;
 };
 
 const TutorialCard = styled(Card)`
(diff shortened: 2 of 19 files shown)


acheroncrypto (commented):
This is a preference, and I don't like this at all. Similar to this, I also don't like setting a value to `undefined`.

`undefined` is the default value in JS, similar to how `()` is the default return value in Rust. Consequently, `return undefined` in JS is similar to `return ()` in Rust. Both seem useless and redundant to me.

rogaldh:
> This is a preference, and I don't like this at all. Similar to this, I also don't like setting a value to `undefined`.
> 
> `undefined` is the default value in JS, similar to how `()` is the default return value in Rust. Consequently, `return undefined` in JS is similar to `return ()` in Rust. Both seem useless and redundant to me.

Why haven't you removed these then https://github.com/solana-playground/solana-playground/blob/master/vscode/tsconfig.json#L12-L14?

acheroncrypto:
That was 4 years ago, so I'm not able to recall why exactly I commented them out instead of removing them. It's likely that I was just using a template/boilerplate from VSCode extension docs that had those options, and I wanted to keep the default settings so that I could recover easily if I ever wanted to go back to the defaults.

You can remove those if you want. The VSCode extension has been unmaintained for a long time though. I'm actually surprised that [it almost has 2k downloads](https://marketplace.visualstudio.com/items?itemName=solana-playground.solpg). This seems like a considerable amount in my eyes, especially considering the Solana developer ecosystem size and the fact that we never even mention its existence anywhere. I think it only had like a couple hundred downloads the last time I checked, which is why I decided it was not worth developing further.

rogaldh:
> That was 4 years ago, so I'm not able to recall why exactly I commented them out instead of removing them. It's likely that I was just using a template/boilerplate from VSCode extension docs that had those options, and I wanted to keep the default settings so that I could recover easily if I ever wanted to go back to the defaults.
> 
> You can remove those if you want. The VSCode extension has been unmaintained for a long time though. I'm actually surprised that [it almost has 2k downloads](https://marketplace.visualstudio.com/items?itemName=solana-playground.solpg). This seems like a considerable amount in my eyes, especially considering the Solana developer ecosystem size and the fact that we never even mention its existence anywhere. I think it only had like a couple hundred downloads the last time I checked, which is why I decided it was not worth developing further.

Got it. I misunderstood your intent then. Thx for the context.

PR #453, opened 2026-05-13
feat(ci): actualize ci/cd workflows to deploy to GAE

rogaldh:
## Context

Server deploys run through GitHub Actions using GCP credentials stored as repo secrets. A `cicd_server.yml` already ran `gcloud app deploy`, but it bundled the format/clippy steps inline, so the checks couldn't be shared with `ci.yml`.

## Problem

- Checks weren't reusable across PR validation and the deploy gate, and the deploy wasn't gated to releases.
- No committed GAE config (`app.yaml`), and the `Dockerfile` wasn't deploy-ready (build args, binary permissions).

## Why

Make the deploy maintainable and safe: reuse the checks across PR and deploy, gate deploys to tagged releases, and get the server image and config deploy-ready.

## Solution

Extract the shared checks into a reusable workflow, keep `ci.yml` as the PR gate, and add a tag-gated `cicd.yml` deploy.

- `reusable-checks.yml` — reusable `workflow_call` bundle of the format/clippy/build jobs; single source of truth for `ci.yml` and `cicd.yml`.
- `ci.yml` — slimmed to a PR-only caller of `reusable-checks.yml`.
- `cicd.yml` — deploy on push to `master` + manual dispatch, but only for a tagged commit: exact-match tag on push, latest tag on manual, fails if none. Version label is the tag with periods → dashes; deployed `--no-promote`.
- `server/app.yaml` — App Engine Flex config sized for concurrent `cargo-build-sbf` builds (`cpu: 4`, `memory_gb: 16`, `PG_BUILD_CONCURRENCY: 4`).
- `server/Dockerfile` — default `USER` arg and `chmod 700` on the server binary.
- `server/README.md` — deployment + versioning docs.
- `docs/local-ci-with-act.md` — running CI locally with `act`.

Traffic promotion stays manual, and `PG_DB_URI` isn't wired by this deploy. Client deployment is handled separately by #447 (Vercel).


.github/workflows/reusable-checks.yml
```diff
@@ -0,0 +1,170 @@
...
+      rust_toolchain:
+        description: "Rust toolchain channel used for `cargo fmt --check` in the server workspace (host triple is resolved automatically by rustup)."
+        required: false
+        type: string
+        default: nightly
+      wasm_rust_toolchain:
+        description: "Rust toolchain channel used for `cargo fmt --check` and `cargo clippy` in the wasm workspace. Pinned because wasm deps (e.g. ahash 0.8.3) reference nightly features removed in newer toolchains."
+        required: false
+        type: string
+        default: "1.75.0"
```
acheroncrypto: Why do we have `rust_toolchain` and `wasm_rust_toolchain` args? Everything already has a dedicated `rust-toolchain.toml` to manage the toolchain. The only exception is the requirement of the
 `nightly` when checking the formatting of the server.

.github/workflows/reusable-checks.yml
```diff
@@ -0,0 +1,170 @@
...
+      # Running under wasm/'s 1.75 pin with --locked sidesteps both, and the install lands at wasm-pack's expected cache path so its existence check short-circuits the fallback.
+      - name: Seed wasm-pack cache with wasm-bindgen-cli
+        working-directory: wasm
+        run: |
+          for v in 0.2.92 0.2.87; do
+            dest="$HOME/.cache/.wasm-pack/wasm-bindgen-cargo-install-$v"
+            cargo install --locked --version "$v" --root "$dest" wasm-bindgen-cli
+            mv "$dest/bin/"* "$dest/"
+            rmdir "$dest/bin"
+          done
```
acheroncrypto: How do you reproduce this issue? We've successfully been building the WASM packages in CI without this code ([example](https://github.com/solana-playground/solana-playground/actions/runs/28070790618/job/83104865065#step:6:13)).

.github/local-ci-with-act.md
acheroncrypto, line 1: Could you move this to the `.github` folder?

server/app.yaml
```diff
@@ -0,0 +1,19 @@
...
+env: flex
+
+automatic_scaling:
+  min_num_instances: 1
+  max_num_instances: 1
+
+# Runtime configuration is supplied via GCP-managed env vars and/or secrets.
+# Keep this file free of credentials so the manifest is safe to commit.
+env_variables:
+  PG_PORT: "8080"
```
acheroncrypto: We can also omit this, given we omitted all other defaults.

```suggestion
```

server/app.yaml
```diff
@@ -0,0 +1,19 @@
...
+  # Each cargo-build-sbf peaks ~3.7GB and ~1 core; sized with resources below for 4 concurrent builds.
+  PG_BUILD_CONCURRENCY: "4"
+
+resources:
+  cpu: 4
+  memory_gb: 16
+  disk_size_gb: 10
```
acheroncrypto: This might be too little for a single instance.

server/README.md
```diff
@@ -22,3 +22,29 @@ This will:
...
+
+The server is deployed to **Google App Engine** as the `playground-server` service via [`.github/workflows/cicd.yml`](../.github/workflows/cicd.yml). The workflow triggers on pushes to `master` and on manual dispatch, but checks and deploy run only for a tagged commit — untagged pushes exit early.
+
+The App Engine version label is the tag with periods replaced by dashes (e.g. `0.29.1` → `0-29-1`) and is uploaded with `--no-promote`, so traffic must be cut over manually in the GCP console.
+
+## Versioning
+
+Tags are the unit of release. The deploy uses `git describe --exact-match`, so the tag must point to the commit at `master` HEAD. To cut a new version:
+
+1. Tag the commit you're releasing (e.g. `git tag 0.29.2`).
```
acheroncrypto: Nit: We've been de-Anchorizing the repo for a while. We'll also support more versions soon™. Could you remove the references to Anchor versions?

acheroncrypto (changes requested):

server/app.yaml
rogaldh: Increased to 50

server/app.yaml
rogaldh: Fair. Will do

server/README.md
rogaldh: Took references from existing tags at the Foundation's fork. Will change

.github/local-ci-with-act.md
rogaldh: Ok. This is still a doc, and I've never seen documentation within `.github` dir, but I assume you do not want to have a folder within the root dir as it is not a service or so.

.github/workflows/reusable-checks.yml
rogaldh: My intent was to pass a toolchain into each cargo call, as different toolchains are used for wasm and for the server. Also, it helps to display the toolchain at the CI, but I agree it could be done by reading logs.
I assume you prefer dropping these, right?

.github/workflows/reusable-checks.yml
rogaldh: I encountered this some time ago and managed to resolve this issue with that part, but currently, I can not reproduce it as well. Will drop this step. Thank you

.github/workflows/reusable-checks.yml
acheroncrypto: Yes. Generally speaking, if it's easy to have a single source of truth, we should prefer that.

.github/local-ci-with-act.md
acheroncrypto: Yes. I have a preference for documenting things where they are used, rather than a dedicated separate place.

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,78 @@
...
+          # Tags are already present from checkout (fetch-depth: 0); no network fetch needed.
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            # Manual deploy: use the latest reachable tag; fail loudly rather than deploy a sentinel version.
+            TAG=$(git describe --tags --abbrev=0 2>/dev/null || true)
+            if [ -z "$TAG" ]; then
+              echo "::error::No git tag found to derive a deploy version; create a tag before dispatching." >&2
+              exit 1
+            fi
+          else
+            TAG=$(git describe --tags --exact-match 2>/dev/null || true)
```
acheroncrypto, line 37: This seems to ignore all errors. It's probably fine, considering the rest of the steps would either be skipped due to an empty string or fail due to an invalid version, but I wanted to note 
just in case this is *not* the intended behavior.

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,78 @@
...
+          fi
+          # Replace periods with dashes for App Engine compatibility (empty stays empty)
+          echo "version_tag=${TAG//./-}" >> "$GITHUB_OUTPUT"
```
acheroncrypto, line 40: Nit: Version tags should have the 'v' prefix e.g. `v0.1.0`, and so the prefix should be stripped here.

Just thinking out loud: I'm not sure if we should even have regular version tags, since we don't actually version anything (except public libs). Should we name tags to something like `v0.0.<YY><MM><DD><C>`? `C` is for the count of tags on that day, so that we can have multiple tags per day (max 10). An example would be `v0.0.2606250` for the first deployment on the 25th of June 2026. You could include an automatic tag creation logic into this workflow too. Wdyt?

server/app.yaml
```diff
@@ -0,0 +1,18 @@
...
+  max_num_instances: 1
+
+# Runtime configuration is supplied via GCP-managed env vars and/or secrets.
+# Keep this file free of credentials so the manifest is safe to commit.
+env_variables:
+  # Each cargo-build-sbf peaks ~3.7GB and ~1 core; sized with resources below for 4 concurrent builds.
+  PG_BUILD_CONCURRENCY: "4"
+
+resources:
+  cpu: 4
```
acheroncrypto, line 16: This should probably be at least 1 higher than `PG_BUILD_CONCURRENCY`, since we'd want the server to still be able to easily handle new requests if all build cores are taken.

Also just FYI, we'll have more control over how many cores and how much memory a build process uses soon™. We can get back to this and `PG_BUILD_CONCURRENCY` after I push those changes.

rogaldh: Probably yes, but I can not say 100%. I wanted to test after the merge to master and increase the limit in the follow-up PR if needed.
Is it okay for you?

.github/workflows/cicd.yml
rogaldh: AFAIK, there is no restriction regarding what should be at the start of the tag. GAE has a requirement that the tag should start and end with either a letter or a digit, and it may contain hyphens.
So digit and letter will work. The current variant allows the use of `v1.2.3` and `1.2.3`. The script is transparent here.

I think that having tags that follow Semver notation is good enough. Yes, Playground does not have tags right now, but nothing prevents us from using it at the fork.
Any automatic solution will rely on a tool, right? I'd rather use an existing tool (like `semantic versioning` or `changeset`), and it will likely use Semver notation already. To bump a patch every time, if needed, seems perfectly aligned with semver.

.github/workflows/cicd.yml
rogaldh: That is intended.
Manual workflow trigger should fail, but casual invocation should pass and not alert with a failure.

Empty tag is handled by following jobs, like so: https://github.com/solana-playground/solana-playground/pull/453/changes/BASE..c3017353a5b436d7ba09dbdf36a32af049501322#diff-6727e33ccc9195d67f0786e1384f8f1cdaf4090c3e77547943105bd2b28c99d0R44
The same condition is on the line 53

.github/workflows/cicd.yml
acheroncrypto: > AFAIK, there is no restriction regarding what should be at the start of the tag. GAE has a requirement that the tag should start and end with either a letter or a digit, and it may contain hyphens.

I see. I'm guessing it doesn't allow periods (`.`)?

> Any automatic solution will rely on a tool, right?

No. Following the convention I described above, we could easily create tags in an automated way with only a couple of lines, since we already know the date and the amount of tags that day.

What I'm saying is, since we don't actually have versions, it probably makes more sense to follow the logic mentioned above. You could also just start from `v0.0.0` and only increase the patch value (e.g. `v0.0.1` -> `v0.0.2`...), but imo the date logic makes versioning clearer.

Both options are SemVer-compatible. If we ever want to switch to regular versioning, we could easily do it. However, currently, regular SemVer doesn't really fit this project because:

- all public usage is intended to stay backwards-compatible (e.g. server routes)
- all internal usage can break at any moment (e.g. client libs)
- not sure if this is still the plan, but I think Jonas mentioned you wanted to deploy each commit

server/app.yaml
acheroncrypto: Yea, that works too. We could probably change the `PG_BUILD_CONCURRENCY` default in server to something like `cores - 1`. This will be more important once each build only takes up 1 core by default. The amount of cores you use is decided by Rust and your OS right now; we don't limit it to a single core.

rogaldh: I'd change the PG_BUILD_CONCURRENCY for now, then. 'cores-1' might be done later.

.github/workflows/cicd.yml
rogaldh: > I see. I'm guessing it doesn't allow periods (`.`)?

Yes

rogaldh: > but imo the date logic makes versioning clearer

The tag triggers workflows. Within the current workflow, the tag is created by a human. Anyone can use what is best suited to them.
I do not see problems here tbh

rogaldh: >No. Following the convention I described above, we could easily create tags in an automated way with only a couple of lines, since we already know the date and the amount of tags that day.

By "relying on a tool", I mean the date thing is custom logic, and most of the tools I saw use auto-incrementing, not the date thing. Do not see problems with incrementing patches.

Also `<YY><MM><DD><C>` has potential issue. YY might have 00 at 2100. I agree that this is a huge horizon, but the pattern has issues. The patch could not start with 0.

acheroncrypto: > The tag triggers workflows. Within the current workflow, the tag is created by a human. Anyone can use what is best suited to them.
> I do not see problems here tbh

The problem is "the tag is created by a human". There is no reason to make people choose the tag when it can easily be automated.

> > No. Following the convention I described above, we could easily create tags in an automated way with only a couple of lines, since we already know the date and the amount of tags that day.
> 
> By "relying on a tool", I mean the date thing is custom logic, and most of the tools I saw use auto-incrementing, not the date thing.

I personally wouldn't call a couple lines of code that we can inline in our workflows "relying on a tool".

>  Do not see problems with incrementing patches.

The problem is that incrementing patches does not correspond to real versioning, even though it looks like it does. This would still be better than letting people choose the tag, but as I mentioned above, date/timestamp would be much clearer imo.

> Also `<YY><MM><DD><C>` has potential issue. YY might have 00 at 2100. I agree that this is a huge horizon, but the pattern has issues. The patch could not start with 0.

OK, if you think we cannot change this tag logic in the next ~74 years, then we could just use the full year `<YYYY>`. I agree that this might be an issue ~8000 years from now, in the year 10,000, but I think we have just enough time to figure that out. Thoughts?

rogaldh: > The problem is "the tag is created by a human". There is no reason to make people choose the tag when it can easily be automated.

I agree with that, but previously you’ve said that you push to master and cleaning the solution up continuously. I do not see how we could make automate solutions around releases with current approach. This PR gonna land to the origin

rogaldh: > I personally wouldn't call a couple lines of code that we can inline in our workflows "relying on a tool".

By “relying on a tool” I mean to add a real solution that other folks use instead of inventing one of ours.

rogaldh: > The problem is that incrementing patches does not correspond to real versioning, even though it looks like it does. This would still be better than letting people choose the tag, but as I mentioned above, date/timestamp would be much clearer imo.

I do not agree with that. Having automated CICD in mind, any code change that is pushed to master produces new variant of the product. It does not matter whether it is “real version” or “not real”. It is a new version with its own changes.

rogaldh: > OK, if you think we cannot change this tag logic in the next ~74 years, then we could just use the full year <YYYY>. I agree that this might be an issue ~8000 years from now, in the year 10,000, but I think we have just enough time to figure that out. Thoughts?

I’m just highlighting a potential problem with proposed variant. I do not see benefits in adding date into the version. Auto incrementing patch is pretty much the same as proposed increment with 10 cap.

Just to clarify, do we talk about automated tag creation at the moment or just the validation of a tag?
Currently at the fork tags are created manually but calling the `git tag`

acheroncrypto: > I do not see how we could make automate solutions around releases with current approach.

I was just referring to the manual tag creation process here. Auto deploying on every change was out-of-scope.

> I do not agree with that. Having automated CICD in mind, any code change that is pushed to master produces new variant of the product. It does not matter whether it is “real version” or “not real”. It is a new version with its own changes.

The problem is that this is not a repository with one product; it's a monorepo with multiple things. Some of these things have their own *proper* versioning. Auto incrementing versions would have made more sense if, for example, we had a dedicated repo for each product.

That being said, I won't die on this hill; you can use the auto-incrementing solution if you want. My main request was to eliminate having to create the tags manually, and the date convention was just for clarity of the fact that it's not "real versioning".

> I do not see benefits in adding date into the version. Auto incrementing patch is pretty much the same as proposed increment with 10 cap.

Using "real versioning" for a thing inside a monorepo that doesn't actually have versioning is not the same thing, but that's alright. We could also just use the full timestamp rather than having 10 cap.

> Just to clarify, do we talk about automated tag creation at the moment or just the validation of a tag? Currently at the fork tags are created manually but calling the `git tag`

Answered above.

rogaldh: Sorry, I'm a bit lost within this thread. Initially, you proposed this:
> "You could include an automatic tag creation logic into this workflow too. Wdyt?"

But later highlighted the opposite:
> "I was just referring to the manual tag creation process here. Auto deploying on every change was out of scope."

To be honest, I lost the moment when we switched to automatic tag creation instead of a manual one. Thanks for clarifying, though.

I think I understand your point about the monorepo and why you're proposing a solution based on dates; I appreciate the patience here. As you mentioned before, this is out of the initial scope; we've got a commitment to Jonas to get the deployment done, and we're running into the timeline.

Would it work for you if we merge the current PR as is to unblock the service deployment? Once that's out, I'll come back to the dates solution with you. That way we hit the deployment commitment and still give the auto-increment logic the attention it needs.

What do you think?

acheroncrypto: > Sorry, I'm a bit lost within this thread. Initially, you proposed this:
> 
> > "You could include an automatic tag creation logic into this workflow too. Wdyt?"
> 
> But later highlighted the opposite:
> 
> > "I was just referring to the manual tag creation process here. Auto deploying on every change was out of scope."

How are they opposite? The former was an idea/suggestion about including automatic tag creation in this workflow. And the latter also talks about:

1. Manual tag creation being an issue, thus implicitly suggesting automatic tag creation again
2. Clarify that automatic tag creation in a workflow file does not mean we'll auto deploy every change.

In this context, "having automatic tag creation" and "not having manual tag creation" point to the same idea. Therefore, they seem consistent to me.

**TL;DR** is to just trigger the deploy workflow, and it handles everything. No manual tag creation.

> Would it work for you if we merge the current PR as is to unblock the service deployment? Once that's out, I'll come back to the dates solution with you. That way we hit the deployment commitment and still give the auto-increment logic the attention it needs.
> 
> What do you think?

Are you going to be deploying from the foundation's fork (including tags)? If so, then that works with me.

rogaldh: > How are they opposite? The former was an idea/suggestion about including automatic tag creation in this workflow. And the latter also talks about:

This is a port of the original PR that allows deployment to GAE. Automatic tag creation was out of scope.
My original question "Any automatic solution will rely on a tool, right?" was about using tools in future to make things automatic. I was referring to the fact that tools use semver tagging, not the custom pattern.

I haven't yet proposed adding automatic things into the current flow.

>Manual tag creation being an issue, thus implicitly suggesting automatic tag creation again
Clarify that automatic tag creation in a workflow file does not mean we'll auto deploy every change.

From my experience, that depends on the practices being chosen within a project. Currently, on the playground, master receives constant updates, and we can not utilise automatic deployments from master. So the tag is proposed as a trigger. Quite a common practice I've seen on a lot of projects.

And it is already used at SF's fork. The tag allows see which version is the latest and has been deployed. That is quite useful. Do you see problems with that?

> Are you going to be deploying from the foundation's fork (including tags)? If so, then that works with me.

Let's merge it then. We sync master at the fork and can not trigger any deploy through GitHub's UI.

server/README.md
```diff
@@ -22,3 +22,29 @@ This will:
 3. Start the server
 
 See the [root README](../README.md#run-with-docker) for more options.
+
+# Deployment
```
acheroncrypto, line 26: This is not the main header.

```suggestion
## Deployment
```

server/README.md
```diff
@@ -22,3 +22,29 @@ This will:
...
+
+The server is deployed to **Google App Engine** as the `playground-server` service via [`.github/workflows/cicd.yml`](../.github/workflows/cicd.yml). The workflow triggers on pushes to `master` and on manual dispatch, but checks and deploy run only for a tagged commit — untagged pushes exit early.
+
+The App Engine version label is the tag with periods replaced by dashes (e.g. `1.2.3` → `1-2-3`) and is uploaded with `--no-promote`, so traffic must be cut over manually in the GCP console.
+
+## Versioning
```
acheroncrypto, line 32:
```suggestion
### Versioning
```

.github/workflows/reusable-checks.yml
```diff
@@ -0,0 +1,149 @@
...
+      - uses: actions/checkout@v6
+
+      # Toolchain + rustfmt come from wasm/rust-toolchain.toml.
+      - name: Check format
+        run: cargo fmt --check
+
+  clippy_server:
+    name: Clippy Server
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
```
acheroncrypto, line 74: This and some others are way too long. This job took 47 seconds in [the last run](https://github.com/solana-playground/solana-playground/actions/runs/29022014746/job/86131439950).
```suggestion
    timeout-minutes: 10
```

.github/workflows/cicd.yml
acheroncrypto: > Currently, on the playground, master receives constant updates, and we can not utilise automatic deployments from master. So the tag is proposed as a trigger. Quite a common practice I've seen on a lot of projects.

My idea was a sort of a "deploy" workflow, similar to how many projects have a "release" workflow that can be triggered on-demand, but this is also fine.

> Let's merge it then.

Merging now.

> We sync master at the fork and can not trigger any deploy through GitHub's UI.

Why can't we trigger the deploy through GitHub's UI?

acheroncrypto (approved):
Just mentioning these nits since I can't commit them to an organization's fork.

Merging to unblock, but it would be great to make it deployable just by triggering the workflow without the manual tag creation.

.github/workflows/cicd.yml
acheroncrypto, line 1: This is a misnomer with how things are currently. It should be named something like `deploy` instead of `cicd`, given we're not continuously deploying. Changed the commit message when merging because of this.

PR #457, opened 2026-05-28
fix(server): clean up env variables to not expose them within builder

rogaldh:
**Context.** The `/build` endpoint spawns `cargo-build-sbf` as a child process at `server/src/program.rs:124` to compile user-submitted Solana programs. Until now the child inherited the full server process env, which on a deployed container includes every `PG_*` config var (e.g. `PG_DB_URI`, `PG_CLIENT_URLS`) plus anything the host passes through.

**Problem.** Any env var present in the server process is visible to user-submitted code at compile time via the `env!` / `option_env!` macros. A submitted program can read e.g. `PG_DB_URI` and bake its value into the compiled artifact, then exfiltrate it through the `stderr` or `idl` fields returned in the build response.

**Why to solve.** The build endpoint accepts untrusted source from the public client. Sensitive deployment configuration (DB credentials, internal URLs, anything an operator puts in env) leaking into compiled programs is a credential-disclosure risk for every operator running the server. The leak is silent — nothing in the current code path warns about it.

**Solution.** Introduce `apply_build_env` at `server/src/program.rs:22` that calls `Command::env_clear()` and then re-injects only the four toolchain-locator vars `cargo-build-sbf` actually needs: `PATH`, `HOME`, `CARGO_HOME`, `RUSTUP_HOME`. Apply at the single subprocess spawn site — a static audit confirmed the server crate has exactly one process-spawn call.
Verified e2e on a docker-compose dev stack: with `PG_CANARY=should-not-leak` set in the server container env, a submitted program containing `env!("PG_CANARY")` fails to compile with `environment variable PG_CANARY not defined`, while `env!("CARGO_HOME")` compiles cleanly.

### DIFF
--- a/server/src/program.rs
+++ b/server/src/program.rs
@@ -16,18 +16,14 @@ const MAX_FILE_AMOUNT: usize = 64;
 const MAX_PATH_LENGTH: usize = 128;
 
 /// Apply a clean environment containing only the toolchain locator vars,
-/// derived from `$HOME` so we don't assume a specific username or container layout.
+/// inherited from the parent process so we don't impose any OS-specific layout.
 fn apply_build_env(cmd: &mut Command) -> &mut Command {
-    let home = std::env::var("HOME").expect("HOME must be set for the build subprocess");
-    let build_path = format!(
-        "{home}/.local/share/solana/install/active_release/bin:{home}/.cargo/bin:\
-         /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-    );
-    cmd.env_clear()
-        .env("PATH", build_path)
-        .env("HOME", &home)
-        .env("CARGO_HOME", format!("{home}/.cargo"))
-        .env("RUSTUP_HOME", format!("{home}/.rustup"));
+    cmd.env_clear();
+    for key in ["PATH", "HOME", "CARGO_HOME", "RUSTUP_HOME"] {
+        if let Ok(val) = std::env::var(key) {
+            cmd.env(key, val);
+        }
+    }
     cmd
 }
 
--- a/server/src/program.rs
+++ b/server/src/program.rs
@@ -15,16 +15,20 @@ const MAX_FILE_AMOUNT: usize = 64;
 /// Maximum length of the file paths to pass to the [`build`] function.
 const MAX_PATH_LENGTH: usize = 128;
 
-/// PATH for the build subprocess.
-const BUILD_PATH: &str = "/home/solpg/.local/share/solana/install/active_release/bin:/home/solpg/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin";
-
-/// Apply a clean environment containing only the toolchain locator vars.
+/// Apply a clean environment containing only the toolchain locator vars,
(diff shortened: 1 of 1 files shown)


server/src/program.rs
```diff
@@ -15,6 +15,18 @@ const MAX_FILE_AMOUNT: usize = 64;
...
+
+/// Apply a clean environment containing only the toolchain locator vars.
+fn apply_build_env(cmd: &mut Command) -> &mut Command {
+    cmd.env_clear()
+        .env("PATH", BUILD_PATH)
+        .env("HOME", "/home/solpg")
+        .env("CARGO_HOME", "/home/solpg/.cargo")
+        .env("RUSTUP_HOME", "/home/solpg/.rustup")
+}
+
```
acheroncrypto: This breaks when the username is not `solpg`. The username is not a requirement. Running the server in a Docker container is also not a requirement. Please either make these values dynamic or filter out the `PG_`  prefixed env variables instead.

acheroncrypto (changes requested):

server/src/program.rs
rogaldh: I'll use HOME to create the PATH. Thx

acheroncrypto: Couldn't we get all the env variables needed dynamically to avoid making OS path assumptions? The server is OS-agnostic currently.

rogaldh: I see, that is fair. Will do

server/src/program.rs
```diff
@@ -15,6 +15,18 @@ const MAX_FILE_AMOUNT: usize = 64;
 /// Maximum length of the file paths to pass to the [`build`] function.
 const MAX_PATH_LENGTH: usize = 128;
 
+/// Apply a clean environment containing only the toolchain locator vars,
+/// inherited from the parent process so we don't impose any OS-specific layout.
+fn apply_build_env(cmd: &mut Command) -> &mut Command {
+    cmd.env_clear();
+    for key in ["PATH", "HOME", "CARGO_HOME", "RUSTUP_HOME"] {
```
acheroncrypto: Are `CARGO_HOME` and `RUSTUP_HOME` needed?

server/src/program.rs
```diff
@@ -109,7 +121,8 @@ pub fn build(
     )?;
 
     // Build the program
-    let output = Command::new("cargo-build-sbf")
+    let mut cmd = Command::new("cargo-build-sbf");
+    let output = apply_build_env(&mut cmd)
```
acheroncrypto: - Log a warning if the env variable doesn't exist (I believe `HOME` is not always defined in all OS')
- Stylistic recommendations:
    - Avoid declaring something as mutable if possible
    - Inline if it's only used once and small*ish*

Something like:

```suggestion
    let output = Command::new("cargo-build-sbf")
        .env_clear()
        .envs(
            ["PATH", "HOME", "CARGO_HOME", "RUSTUP_HOME"]
                .into_iter()
                .filter_map(|key| {
                    std::env::var(key)
                        .inspect_err(|e| warn!("Failed to get env variable: `{key}`: {e}"))
                        .ok()
                        .map(|value| (key, value))
                }),
        )
```

server/src/program.rs
rogaldh: Yes. In the latest state they needed no more

acheroncrypto (approved):
Thank you!

PR #465, opened 2026-07-24
Add ability to scroll past end in test UI

### LINKED ISSUE #168: Add ability to scroll past end in test UI
### Problem

Test UI could get quite long as it's autogenerated based on the IDL and this results in excessive scrolling to the bottom during expansions.

### Solution

Implement scroll-past-end feature i.e. the user should be able to scroll down to put the bottom component in test UI to the top similar to how editors like `monaco-editor` behave.

0xfandom:
### Problem

The test UI is autogenerated from the IDL and can get long. Expanding instructions/accounts near the bottom forces excessive scrolling, and the last component can't be brought up into comfortable view.

### Change

Add `padding-bottom` to the test panel's wrapper so the view can scroll past the end — the last component can be scrolled to the top, matching the scroll-beyond-last-line behavior of editors like monaco-editor.

Scoped to the test panel wrapper so other sidebar pages are unaffected.

Closes #168

### DIFF
--- a/client/src/views/sidebar/test/Component/Test.tsx
+++ b/client/src/views/sidebar/test/Component/Test.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import BN from "bn.js";
 
 import Account from "./Account";
@@ -138,11 +138,24 @@ const InitialWrapper = styled.div`
 `;
 
 const Wrapper = styled.div`
-  user-select: none;
-  display: flex;
-  flex-direction: column;
-  gap: 1rem;
-  padding-top: 1rem;
+  ${({ theme }) => css`
+    user-select: none;
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+    padding-top: 1rem;
+    /**
+     * Allow scrolling past the end so the last component can reach the top of
+     * the view, similar to the "scroll beyond last line" behavior in editors.
+     */
+    padding-bottom: calc(
+      100vh -
+        (
+          ${theme.views.sidebar.right.title.height} +
+            ${theme.views.bottom.height}
+        )
+    );
+  `}
 `;
 
 const ProgramNameWrapper = styled.div`


javi0guerrero4-cloud (commented):
Fix

client/src/views/sidebar/test/Component/Test.tsx
```diff
@@ -143,6 +143,9 @@ const Wrapper = styled.div`
   flex-direction: column;
   gap: 1rem;
   padding-top: 1rem;
+  /* Allow scrolling past the end so the last component can reach the top of the
+     view, similar to the "scroll beyond last line" behavior in editors. */
```
acheroncrypto: Nit: repo preference:


```suggestion
  /**
   * Allow scrolling past the end so the last component can reach the top of the
   * view, similar to the "scroll beyond last line" behavior in editors.
   */
```

client/src/views/sidebar/test/Component/Test.tsx
```diff
@@ -143,6 +143,9 @@ const Wrapper = styled.div`
...
+  padding-bottom: 100vh;
```
acheroncrypto: I don't think this is correct because sidebar content does not take full view height. There is also the sidebar title at the top and the bottom bar at the bottom currently. Luckily, we already have the height of those views available. See similar usage here:

https://github.com/solana-playground/solana-playground/blob/cd5555155c61572c8c49fb351890519af9e493ef/client/src/components/Wallet/Wallet.tsx#L83-L89

You can get `theme` like this:

https://github.com/solana-playground/solana-playground/blob/cd5555155c61572c8c49fb351890519af9e493ef/client/src/components/Wallet/Wallet.tsx#L76-L77

Note that the `bottom.default.height` is actually redundant. It's a left-over structure from a previous version. I've just realized it while writing this. I'll fix that this week. You can use `bottom.height` instead from now on:

```suggestion
  padding-bottom: calc(100vh - ${theme.views.sidebar.right.title.height} + ${theme.views.bottom.height});;
```

acheroncrypto (changes requested):

0xfandom:
Thanks for the review. Updated:

- Switched to the theme heights instead of `100vh` and pulled `theme` into the styled block.
- Applied the comment style nit.

One small deviation: I used `theme.views.bottom.default.height` rather than `bottom.height`, since `bottom.height` isn't defined in the theme yet (only `bottom.default.height` is set). Happy to switch to `bottom.height` once your cleanup lands.

client/src/views/sidebar/test/Component/Test.tsx
```diff
@@ -138,11 +138,21 @@ const InitialWrapper = styled.div`
...
+    flex-direction: column;
+    gap: 1rem;
+    padding-top: 1rem;
+    /**
+     * Allow scrolling past the end so the last component can reach the top of
+     * the view, similar to the "scroll beyond last line" behavior in editors.
+     */
+    padding-bottom: calc(
+      100vh - ${theme.views.sidebar.right.title.height} +
+        ${theme.views.bottom.default.height}
```
acheroncrypto: > One small deviation: I used `theme.views.bottom.default.height` rather than `bottom.height`, since `bottom.height` isn't defined in the theme yet (only `bottom.default.height` is set). Happy to switch to `bottom.height` once your cleanup lands.

Please switch; it landed yesterday (https://github.com/solana-playground/solana-playground/commit/08a26da0e0f8e5322f485fb1753db673dd25be35).


```suggestion
        ${theme.views.bottom.height}
```

acheroncrypto: We must subtract both the sidebar title height and bottom height:

```suggestion
      100vh - (${theme.views.sidebar.right.title.height} +
        ${theme.views.bottom.height})
```

(sorry my initial snippet didn't have the parantheses)

acheroncrypto (changes requested):

0xfandom:
Both addressed, and rebased on latest `master`:

- Switched to `theme.views.bottom.height` now that 08a26da has landed.
- Subtracting both the sidebar title height and the bottom height.

One note on the snippet: applying it verbatim fails `yarn check-format`, since Prettier reformats the `calc()` when the subtraction wraps. Kept your parenthesised expression and let Prettier lay it out:

```css
padding-bottom: calc(
  100vh -
    (
      ${theme.views.sidebar.right.title.height} +
        ${theme.views.bottom.height}
    )
);
```

Same result, just Prettier-clean. Happy to shape it differently if you prefer another form.

acheroncrypto (approved):
Thanks!

PR #466, opened 2026-07-24
Automatically fetch on-chain IDL when a program is deployed

### LINKED ISSUE #93: Automatically fetch on-chain IDL
Playground already generates UI for interacting with programs that have Anchor IDL(either built inside playground or uploaded the file manually) but it doesn't fetch on-chain IDLs.

Playground should check the program id to see if there is an Anchor IDL on-chain. This not only would allow anyone to easily interact with all programs with on-chain IDL without manually uploading it to playground, but since playground also works with localhost, it would also make the local development experience smoother.

0xfandom:
### Problem

When a user sets a custom program id (e.g. to interact with a pre-deployed program), the playground has no way to obtain the program's Anchor IDL. The IDL is only ever populated from a local build or a manual JSON import, so testing a pre-deployed program requires manually importing its IDL.

### Change

Add an effect that automatically fetches the on-chain IDL when a custom program id is set, reusing the existing `PgProgramInfo.fetchIdl`.

- Limited to custom program ids — a program built in the playground already gets its IDL from the build output.
- Never overwrites an existing IDL (imported or built), and re-checks state after the async fetch so newer data isn't clobbered.
- Runs in the background; fetch failures (connection not ready / no on-chain IDL) are ignored.

This matches the approach suggested in the issue thread (fetch once a custom program id is set).

Closes #93

### DIFF
--- a/client/src/effects/fetch-idl/fetch-idl.ts
+++ b/client/src/effects/fetch-idl/fetch-idl.ts
@@ -1,0 +1,41 @@
+import { PgCommon, PgFramework, PgProgramInfo, PgTerminal } from "../../utils";
+
+/**
+ * Automatically fetch the on-chain Anchor IDL of the current program.
+ *
+ * This runs when the program is deployed and there is no IDL available
+ * locally, e.g. when interacting with a pre-deployed program via a custom
+ * program id or an imported keypair. An existing IDL (imported or built) is
+ * never overwritten.
+ */
+export const fetchIdl = () => {
+  return PgCommon.batchChanges(async () => {
+    const programId = PgProgramInfo.pk;
+
+    // Only fetch when the program is deployed and there is no local IDL.
+    if (!programId || PgProgramInfo.idl || !PgProgramInfo.onChain?.deployed) {
+      return;
+    }
+
+    // TODO: Remove once `PgProgramInfo.fetchIdl` is framework-agnostic
+    // (Seahorse programs are Anchor-compatible and also have Anchor IDLs)
+    const framework = await PgFramework.getFromFiles();
+    if (framework?.name !== "Anchor" && framework?.name !== "Seahorse") return;
+
+    try {
+      const result = await PgProgramInfo.fetchIdl(programId);
+      if (!result) return;
+
+      // Bail if the state changed while fetching (the IDL was set or the
+      // program id changed) so newer data is not overwritten.
+      if (PgProgramInfo.idl || !PgProgramInfo.pk?.equals(programId)) return;
+
+      PgProgramInfo.update({ idl: result.idl });
+      PgTerminal.println(PgTerminal.success("Fetched the on-chain IDL."));
+    } catch {
+      // Fetching can fail for various reasons, e.g. the connection not being
+      // ready or the request getting rate-limited. Ignore because this runs
+      // automatically in the background.
+    }
+  }, [PgProgramInfo.onDidChangePk, PgProgramInfo.onDidChangeOnChain]);
+};
--- a/client/src/effects/fetch-idl/index.ts
+++ b/client/src/effects/fetch-idl/index.ts
@@ -1,0 +1,1 @@
+export * from "./fetch-idl";


client/src/effects/fetch-idl/fetch-idl.ts
```diff
@@ -0,0 +1,34 @@
+import { PgCommon, PgConnection, PgProgramInfo } from "../../utils";
+
+/**
+ * Automatically fetch the on-chain Anchor IDL when the user sets a custom
+ * program id.
+ *
+ * This is limited to custom program ids because a program built in the
+ * playground already gets its IDL from the build output. An existing IDL
+ * (imported or built) is never overwritten.
```
acheroncrypto: Thanks for following the approach mentioned in the original issue (https://github.com/solana-playground/solana-playground/issues/93#issuecomment-1345355299). However, I have to say I don't really remember why I specifically said custom ID there. My guess would be to avoid redundant requests.

Building in the playground is not the only way users can have a program keypair; users can also import a full keypair for the program. I currently don't see any problem fetching the IDL of such programs.

client/src/effects/fetch-idl/fetch-idl.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const fetchIdl = () => {
+  return PgCommon.batchChanges(async () => {
+    const programId = PgProgramInfo.customPk;
```
acheroncrypto: In line with the previous comment, this should be `PgProgramInfo.pk` instead.
```suggestion
    const programId = PgProgramInfo.pk;
```

client/src/effects/fetch-idl/fetch-idl.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // Only fetch for a custom program id and never overwrite an existing IDL.
+    if (!programId || PgProgramInfo.idl) return;
```
acheroncrypto: Can also check if the program itself has been deployed with `PgProgramInfo.onChain?.deployed`.

```suggestion
    if (!programId || PgProgramInfo.idl || !PgProgramInfo.onChain?.deployed) return;
```

acheroncrypto (changes requested):
It's mostly looking good, but it would be nice to also account for the scenario in which the current framework is not Anchor. This is because `PgProgramInfo.fetchIdl` currently only works with Anchor. The plan is to make that function framework-agnostic soon™, but in the meantime we could get the current framework from `PgFramework.getFromFiles()` (passing no args works). Please also add a comment to remove the check:

```ts
// TODO: Remove once `PgProgramInfo.fetchIdl` is framework-agnostic`
```

0xfandom:
Thanks, all addressed:

- Uses `PgProgramInfo.pk` now, so imported-keypair programs are covered too.
- Added the `PgProgramInfo.onChain?.deployed` guard.
- Gated on the current framework being Anchor via `PgFramework.getFromFiles()`, with the `// TODO: Remove once PgProgramInfo.fetchIdl is framework-agnostic` note.

Also updated the change subscriptions to `onDidChangePk` + `onDidChangeOnChain` so it re-runs on deploy/connection changes.

acheroncrypto (commented):
Thanks! It's looking good now. I only left a couple nits. Could you also update the PR title since it's outdated now (no longer only for *custom program ids*)?

client/src/effects/fetch-idl/fetch-idl.ts
```diff
@@ -0,0 +1,38 @@
...
+    const programId = PgProgramInfo.pk;
+
+    // Only fetch when the program is deployed and there is no local IDL.
+    if (!programId || PgProgramInfo.idl || !PgProgramInfo.onChain?.deployed) {
+      return;
+    }
+
+    // TODO: Remove once `PgProgramInfo.fetchIdl` is framework-agnostic
+    const framework = await PgFramework.getFromFiles();
+    if (framework?.name !== "Anchor") return;
```
acheroncrypto: Nit: Although abandoned, `Seahorse` also uses Anchor IDLs.

client/src/effects/fetch-idl/fetch-idl.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (!result) return;
+
+      // Bail if the state changed while fetching (the IDL was set or the
+      // program id changed) so newer data is not overwritten.
+      if (PgProgramInfo.idl || !PgProgramInfo.pk?.equals(programId)) return;
+
+      PgProgramInfo.update({ idl: result.idl });
+    } catch {
+      // Fetching can fail if the connection is not ready or the program has no
+      // on-chain IDL. Ignore because this runs automatically in the background.
```
acheroncrypto: Nit: There are more ways this can fail (e.g. getting rate-limited), and program having no on-chain IDL is not an error case:

https://github.com/solana-playground/solana-playground/blob/223861be2e9b7579a4b9608459b265a04bb9dc5f/client/src/utils/program-info.ts#L212-L214

0xfandom:
Thanks! Nits addressed and rebased on latest `master`:

- Seahorse now passes the framework check too, since it produces Anchor IDLs. Unknown/other frameworks still skip.
- Reworded the `catch` comment: the no-on-chain-IDL case isn't an error (`fetchIdl` returns `null` there and it's handled above), so the comment now points at the actual failure modes, e.g. the connection not being ready or the request getting rate-limited.

Also updated the PR title to match.

acheroncrypto (commented):
Thanks for updating! It's looking great.

One more potential improvement: there is currently no indication when the IDL is automatically fetched. It would be nice to notify the user on success using [`PgTerminal.println`](https://github.com/solana-playground/solana-playground/blob/9d816353a123fc6c4f10edf8679d38bd617a7d96/client/src/utils/terminal/terminal.ts#L72). Something like:

https://github.com/solana-playground/solana-playground/blob/9d816353a123fc6c4f10edf8679d38bd617a7d96/client/src/commands/airdrop/airdrop.ts#L81

0xfandom:
Good call, added:

```ts
PgProgramInfo.update({ idl: result.idl });
PgTerminal.println(PgTerminal.success("Fetched the on-chain IDL."));
```

Only on the success path, so the silent-skip cases (no IDL on chain, non-Anchor framework, fetch failure) stay quiet. Also rebased on latest `master`.

Happy to reword the message if you'd prefer something else.

acheroncrypto (approved):
This is pretty much exactly the same as how I would write this code. Great contribution, thank you!

PR #467, opened 2026-07-29
chore: allowlist Vercel client origins on the server; carry over #447 review follow-ups

rogaldh:
**Context.** The GAE server's CORS allowlist (`PG_CLIENT_URLS`, prefix-matched) still runs on its compiled-in default, which predates the Vercel client deployment from #447. Bundled: the remaining CI/doc follow-ups from the #447 post-merge review.

**Problem.** CORS rejects every request from the Vercel client — production (`solana-playground-two.vercel.app`) and preview (`solana-playground-<hash>-solana-foundation.vercel.app`) alike. From the review: client tests are not run in CI, and the deploy doc pins a Node version that duplicates `package.json` `engines`.

**Why to solve.** The deployed Vercel client cannot build or deploy programs against the foundation-hosted server; the rest closes out the review asks left open at merge.

Notes: the `solana-playground-` prefix entry admitting dynamic preview hostnames is temporary and may be dropped later for a stable alias domain. Verified against a locally-run server — real deployment URLs pass preflight, unrelated origins are blocked.

### DIFF
--- a/server/app.yaml
+++ b/server/app.yaml
@@ -11,6 +11,8 @@ automatic_scaling:
 env_variables:
   # Each cargo-build-sbf peaks ~3.7GB and ~1 core; sized with resources below for 3 concurrent builds, leaving 1 core for serving.
   PG_BUILD_CONCURRENCY: "3"
+  # Prefix-matched against Origin (no trailing slashes). The `solana-playground-` prefix temporarily admits dynamic Vercel preview URLs and might be dropped later.
+  PG_CLIENT_URLS: "https://beta.solpg.io,https://solana-playground-two.vercel.app,https://solana-playground-,http://localhost"
 
 resources:
   cpu: 4
--- a/.github/workflows/reusable-checks.yml
+++ b/.github/workflows/reusable-checks.yml
@@ -145,5 +145,12 @@ jobs:
       - name: Check git status
         run: git status --porcelain | wc -l | grep -q -w "0"
 
+      # Unlike the build, `tsc --noEmit` also covers test files.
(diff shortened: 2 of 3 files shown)


server/app.yaml
```diff
@@ -11,6 +11,8 @@ automatic_scaling:
 env_variables:
   # Each cargo-build-sbf peaks ~3.7GB and ~1 core; sized with resources below for 3 concurrent builds, leaving 1 core for serving.
   PG_BUILD_CONCURRENCY: "3"
+  # Prefix-matched against Origin (no trailing slashes). The `solana-playground-` prefix temporarily admits dynamic Vercel preview URLs and might be dropped later.
+  PG_CLIENT_URLS: "https://beta.solpg.io,https://solana-playground-two.vercel.app,https://solana-playground-,http://localhost"
```
acheroncrypto, line 15: Nit: `https://solana-playground-` already covers `https://solana-playground-two.vercel.app`

acheroncrypto (approved):
You're probably already aware of the comment. It's not a big deal. Thanks!

server/app.yaml
rogaldh: Preview versions have this pattern: `solana-playground-<hash>-solana-foundation.vercel.app` and that is not covered by https://solana-playground-two.vercel.app

acheroncrypto: I didn't say `https://solana-playground-two.vercel.app` covers anything; I said `https://solana-playground-` covers it (and the rest). In other words, having `https://solana-playground-two.vercel.app` is redundant.

PR #469, opened 2026-08-02
docs: fix nonexistent Docker Compose profile in client README

latent-9:
`client/README.md` documents running the standalone client with:

```sh
docker compose -f ../compose.yaml --profile standalone up --build
```

But `compose.yaml` defines no profile named `standalone` — the profiles are `dev`, `prod`, and `client-standalone`. `--profile standalone` matches zero services, so the command starts nothing. The correct profile is **`client-standalone`** (`compose.yaml`), which the root `README.md` already uses (`docker compose up client-standalone`). Fixed the profile name in both the prose and the command in `client/README.md`.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -26,10 +26,10 @@ docker compose -f ../compose.yaml --profile dev up --build
 
 ### Standalone (client only, without the server)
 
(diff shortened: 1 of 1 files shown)


acheroncrypto (approved):
Thanks!

PR #472, opened 2026-08-07
Suggest solutions for common build errors

### LINKED ISSUE #312: InitSpace not found for struct
error[E0599]: no associated item named `InitSpace` found for struct `Favorites` in the current scope
  --> src/lib.rs:54:56
   |
35 | pub struct Favorites{
   | -------------------- associated item `InitSpace` not found for this struct
...
54 |         space = ANCHOR_DISCRIMINATOR_SIZE + Favorites::InitSpace,
   |                                                        ^^^^^^^^^ associated item not found in `Favorites`
 
While completing project 1 in the bootcamp tutorial, I came across this issue. I'm not quite sure how to fix it

0xfandom:
### Problem

Some build errors have a well-known solution that isn't obvious from the compiler output alone. The one in #312:

```
error[E0599]: no associated item named `InitSpace` found for struct `Favorites` in the current scope
```

The constant that's actually wanted is `INIT_SPACE`, and it only exists after adding `#[derive(InitSpace)]` to the struct. Neither is discoverable from the error.

From your comment on the issue:

> Not closing this issue because we might be able to suggest the above solution when you get this compile error.

### Change

Add a `SUGGESTIONS` list that maps an error pattern to a solution, and append the matching suggestions to the build output:

```
error[E0599]: no associated item named `InitSpace` found for struct `Favorites` in the current scope
  --> src/lib.rs:54:56
   |
...
Suggestion: The constant is named `INIT_SPACE`, not `InitSpace`. It's generated by adding `#[derive(InitSpace)]` to the struct.
```

Two patterns are covered, since the mistake shows up in both directions:

- `InitSpace` not found — the derive macro's name used as if it were the constant
- `INIT_SPACE` not found — correct name, but the struct isn't deriving `InitSpace`

Only the errors that are actually shown are checked, as suggesting a solution for an error that got cut by `MAX_ERROR_AMOUNT` would be confusing.

This lives inside the existing `build.improveErrors` block, so it's covered by the "Improve build errors" setting.

Closes #312.

### Notes

- The list is intentionally easy to extend — if there are other errors you'd like covered, I'm happy to add them here.
- I reworded the `displayErrors` slice into a `shownErrors` variable so the suggestions and the output are guaranteed to be derived from the same set of errors.
- Verified with `tsc --noEmit` and `yarn check-format`.


Trustinbtc999-hue:
add

Trustinbtc999-hue (approved):

client/src/commands/build/build.ts
```diff
@@ -333,10 +354,18 @@ const improveOutput = (output: string) => {
...
+        output = shownErrors.reduce((acc, cur) => acc + cur, "");
+
+        // Suggest solutions for common errors. Only the shown errors are
+        // checked because suggesting a solution for an error that is not in
+        // the output would be confusing.
+        const suggestions = SUGGESTIONS.filter((s) => {
+          return shownErrors.some((err) => s.pattern.test(err));
+        });
+        for (const { suggestion } of suggestions) {
+          output += `${PgTerminal.bold("Suggestion:")} ${suggestion}\n`;
```
acheroncrypto: "Suggestion:" gets automatic highlighting:

- https://github.com/solana-playground/solana-playground/blob/fdbf80401add07fc023ecc6863d18fb5a3cea3b6/client/src/utils/terminal/tty.ts#L451
- https://github.com/solana-playground/solana-playground/blob/fdbf80401add07fc023ecc6863d18fb5a3cea3b6/client/src/utils/terminal/tty.ts#L470

```suggestion
          output += `Suggestion: ${suggestion}\n`;
```

acheroncrypto (changes requested):
Small comment; looks good otherwise.

0xfandom:
Applied, and rebased on latest `master`.

Good catch — the manual highlighting was actively working against itself. Wrapping it meant the line no longer started with `Suggestion:`, so the `^(.*?:)` rule in `tty.ts` couldn't match it in the first place.

acheroncrypto (approved):
Thank you!

PR #476, opened 2026-09-22
feat(lsp): add rust analyzer lsp

rogaldh:
## TLDR

Adds an opt-in **Server** backend for Rust intellisense that runs the real rust-analyzer on the
build template's toolchain, alongside the existing in-browser WASM one. Completions and
`cargo check` errors then match the crate versions the build actually uses. WASM stays the default.

## Problem

The editor's rust-analyzer runs as WASM in a browser worker against a fixed set of bundled crates.
It cannot run `cargo check`, and it does not know the dependency versions a build uses. Its
diagnostics drift from what the build server compiles: a type error the build reports may not show
in the editor, and a completion the editor offers may not resolve in the build.

## Context

The playground already builds programs server-side, in per-template Docker images that carry the
exact toolchain and crates. The WASM analyzer is client-only and cannot see any of that. The server
already picks a build template by byte-matching a project's cargo files, but the client never sent
them, so every unstable build used the default template.

## Proposed solution

Run the real rust-analyzer inside the template's program image and bridge it to the editor over a
WebSocket, selected by a setting.

- A **Rust Analyzer** setting picks the backend: `Browser (WASM)` (default) or `Server` (needs a
  server started with `--features unstable`).
- Client LSP stack under `.../rust/lsp/`: JSON-RPC over a WebSocket (`vscode-ws-jsonrpc`), a
  hand-rolled LSP type subset, LSP-to-Monaco converters, Monaco providers, path mapping, status,
  and session orchestration. The Monaco glue is hand-rolled because `monaco-languageclient` needs
  the `@codingame/monaco-vscode-api` stack across the whole editor.
- Server bridge: `server/src/lsp.rs` (hardened session container plus `Content-Length` framing) and
  `routes/unstable/lsp.rs` (WS route, origin check, JSON-RPC dispatch). `solpg/open` ships the
  files and picks the template; `solpg/sync` writes `src/` to disk before `didSave`, as
  `cargo check` reads the disk.
- Template selection shared between `unstable/build` and `unstable/lsp`: `server/src/program.rs`
  validates project files (count, size, path shape) and `Template::find` picks the template from
  the root `Cargo.toml` alone. A `Cargo.lock`, when sent, is byte-checked too. The client now sends
  both root cargo files to the unstable build route, so build and LSP agree on the template.
- Session limits are env-configurable: `PG_LSP_CONCURRENCY`, `PG_LSP_CPU_LIMIT`,
  `PG_LSP_MEMORY_LIMIT`, `PG_LSP_IDLE_TIMEOUT`, `PG_LSP_MAX_LIFETIME` (documented in
  `server/README.md`). `PG_PAYLOAD_LIMIT` caps project bytes; the origin check reuses the
  `PG_CLIENT_URLS` allowlist.
- Bottom-bar `rust-analyzer` indicator, clickable to reconnect.
- Anchor project defaults (root `Cargo.toml`, no vendored `Cargo.lock`) so new Anchor projects
  match `anchor-1.1.2`; `Dockerfile.program` adds the `rust-analyzer` and `rust-src` components
  plus a warm `cargo check`.

Left out on purpose: no per-user gate beyond the session limits above. On Apple Silicon a cold
image build needs `DOCKER_DEFAULT_PLATFORM=linux/amd64`, since Solana ships x86_64 Linux binaries
only; documented in `server/README.md`.

## Verification

- `yarn test-types`, `yarn check-format`, `yarn test-unit` (33 passing, 6 suites): clean.
- `cargo test --features unstable` (5 passing), `cargo fmt --check`: pass.
- Browser: WASM intellisense on a stable server; server LSP on both templates (a project without a
  root `Cargo.toml` selects `program-legacy`, a new Anchor project selects `program-anchor-1.1.2`),
  with template-aware completion, reconnect, and a red indicator when the server stops.

## Checklist

- [x] CI/CD checks pass on the PR
- [ ] A program `build` and `deploy` through a server-backed session


server/images/Dockerfile.program
```diff
@@ -60,6 +63,10 @@ COPY --chown=${USER}:${USER} templates/${TEMPLATE}/ .
 ARG INITIAL_BUILD_COMMAND
 RUN sh -c "${INITIAL_BUILD_COMMAND}"
 
+# Warm the host-target check the language server runs: pre-builds the
+# proc-macro crates, so the first diagnostics take seconds instead of minutes
+RUN cargo check --locked
```
acheroncrypto: It's already "warm" after the `INITIAL_BUILD_COMMAND` above (that's the whole point of it). This is running `cargo check` on a target we don't build for (the default Rust target). This means potentially incorrect results and redundant storage and wasted compute on the server.

Instead, it should use the correct target `--target sbf-solana-solana`. We might need to include the target name in the templates, as they've renamed this a couple times in the past.

server/src/middlewares/cors.rs
```diff
@@ -5,16 +5,23 @@ use tower_http::cors::{AllowOrigin, CorsLayer};
...
+/// Whether requests from `origin` are allowed.
+///
+/// Prefix matching, because `client_urls` entries like `https://solana-playground-`
+/// stand for every Vercel preview deployment.
+pub fn is_allowed_origin(origin: &[u8], client_urls: &[String]) -> bool {
+    client_urls
+        .iter()
+        .any(|url| origin.starts_with(url.as_bytes()))
+}
+
```
acheroncrypto: Nit: Move this below `cors`:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/CONTRIBUTING.md#L46

server/src/routes/unstable/lsp.rs
```diff
@@ -0,0 +1,312 @@
...
+const END_CLIENT_CLOSED: &str = "client closed";
+const END_SERVER_EXITED: &str = "language server exited";
+const END_IDLE_TIMEOUT: &str = "idle timeout";
+const END_MAX_LIFETIME: &str = "max lifetime";
+
+/// Language server state shared between sessions
+#[derive(Clone)]
+pub struct LspState {
+    /// Limits the number of concurrent sessions
+    sem: Arc<Semaphore>,
```
acheroncrypto: This is simply a middleware now:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/main.rs#L52

```suggestion
```

server/src/templates/mod.rs
```diff
@@ -97,17 +99,54 @@ impl Template {
         image_build_args
     }
 
+    /// Pick the template from a project's `cargo` files, the default one when
+    /// there are none.
+    ///
+    /// The manifest selects it; a new Anchor project sends one with no
+    /// `Cargo.lock`, so selection must work from the manifest alone.
```
acheroncrypto: 1. Why does an LSP impl change how templates are decided?
2. Why does a general template implementation, which has no idea what Anchor is, being changed based on an implementor?

Please remove all Anchor references from this file.

server/src/templates/mod.rs
```diff
@@ -97,17 +99,54 @@ impl Template {
...
+                for template in get_all_templates() {
+                    if template.matches(manifest, lock)? {
+                        return Ok(template);
+                    }
+                }
+                Err(anyhow!(
+                    "The `cargo` files match no build template: the dependency set \
+                    is fixed by the build images. Revert `Cargo.toml` to restore \
+                    builds and intellisense"
+                ))
```
acheroncrypto: This is a client-side job:

```suggestion

                Err(anyhow!("The `cargo` files match no build template"))
```

server/src/templates/mod.rs
```diff
@@ -97,17 +99,54 @@ impl Template {
...
     /// Get whether the given cargo files matches the template.
-    pub fn matches(&self, manifest: &str, lock: &str) -> Result<bool> {
+    ///
+    /// The lock is optional to support projects that send a manifest without
+    /// one: a new Anchor project no longer bundles a `Cargo.lock`, so the LSP
+    /// `open` (and the build) must select a template from the manifest alone.
+    /// That is sound because the manifest is the discriminator: it is distinct
+    /// per template and selects the build image, which fixes the exact
+    /// versions. A lock, when supplied, is byte-checked too.
+    pub fn matches(&self, manifest: &str, lock: Option<&str>) -> Result<bool> {
```
acheroncrypto: Incorrect: on the contrary, Anchor has always tried to avoid "bundling" `Cargo.lock` (or any other lock file). It only started including it ~2 weeks back (https://github.com/otter-sec/anchor/pull/4684#discussion_r3989736428) to fix `litesvm`-related dependency issues.

Additionally, whether Anchor, or any other framework, bundles a lock file or not does not concern us; playground builds will require `Cargo.lock` in the foreseeable future to keep the builds reliable and reproducible. Clients also need this information so that they can export the project without running into dependency-related problems.

server/src/config.rs
```diff
@@ -25,6 +25,16 @@ pub struct Config {
...
+    /// Maximum amount of concurrent language server sessions
+    pub lsp_concurrency: usize,
+    /// CPU (cores) limit of a language server session container
+    pub lsp_cpu: usize,
+    /// Memory limit of a language server session container in bytes
+    pub lsp_memory: usize,
+    /// Seconds without a client message before a language server session is closed
+    pub lsp_idle_timeout: u64,
+    /// Seconds after which a language server session is closed regardless of activity
+    pub lsp_max_lifetime: u64,
```
acheroncrypto: Please check out the existing unstable configurations and make this similar to them (a single field):

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/config.rs#L24-L27

server/src/config.rs
```diff
@@ -86,6 +96,11 @@ impl Config {
                     },
                 },
             },
+            lsp_concurrency: get_env("LSP_CONCURRENCY", 4usize),
```
acheroncrypto: This means 4 people in total can use RA at the same time?

server/src/config.rs
```diff
@@ -86,6 +96,11 @@ impl Config {
...
+            lsp_cpu: get_env("LSP_CPU_LIMIT", 1usize),
+            lsp_memory: get_env("LSP_MEMORY_LIMIT", 4usize * 1024 * 1024 * 1024),
```
acheroncrypto: This is too high. Unstable builds currently default to 2 GiB:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/config.rs#L56-L59

The release build is much more expensive than running `cargo check`. It should be *at most* the same value as the build default. I'd go with 1 GiB as a starter. It could even be lower but needs testing.

server/src/lsp.rs
```diff
@@ -0,0 +1,380 @@
+//! Long-lived language server sessions inside the template images.
+//!
+//! A build runs a container for seconds; a language server session lives as
+//! long as the editor is open and streams stdio, which is why this is a
+//! separate small type rather than a mode of [`crate::Sandbox`].
```
acheroncrypto: Wdyt about adding a `spawn` method to `Sandbox`? This file is already a bit out-of-sync with the `Sandbox` impl.

server/src/lsp.rs
```diff
@@ -0,0 +1,380 @@
...
+        Self {
+            cpu: 1,
+            // 4 GiB: `cargo check` on an Anchor program
+            memory: 4 * 1024 * 1024 * 1024,
+            idle_timeout: Duration::from_secs(10 * 60),
+            max_lifetime: Duration::from_secs(4 * 60 * 60),
+            max_files_bytes: 1024 * 1024,
+        }
+    }
+}
```
acheroncrypto: Is this being used?

Also, the defaults should *only* be set in [the `config.rs` file](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/config.rs).

server/src/lsp.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub max_lifetime: Duration,
+    /// Total bytes of project files accepted in one `open`/`sync`
+    pub max_files_bytes: usize,
+}
+
+impl Default for Limits {
+    fn default() -> Self {
+        Self {
+            cpu: 1,
+            // 4 GiB: `cargo check` on an Anchor program
```
acheroncrypto: Which Anchor program? Again, there is no need to mention Anchor here.

server/src/main.rs
```diff
@@ -58,6 +61,20 @@ async fn main() -> Result<()> {
...
+                    config.lsp_concurrency,
+                    solpg_server::lsp::Limits {
+                        cpu: config.lsp_cpu,
+                        memory: config.lsp_memory,
+                        idle_timeout: Duration::from_secs(config.lsp_idle_timeout),
+                        max_lifetime: Duration::from_secs(config.lsp_max_lifetime),
+                        max_files_bytes: config.payload_limit,
+                    },
+                    config.client_urls.clone(),
+                )),
```
acheroncrypto: Please follow the existing conventions:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/main.rs#L51

server/src/main.rs
```diff
@@ -58,6 +61,20 @@ async fn main() -> Result<()> {
                     .with_state(unstable::BundleState::new(config.unstable_bundle))
                     .layer(concurrency_limit(bundle_concurrency)),
             )
+            .route(
+                "/lsp",
+                get(unstable::lsp).with_state(unstable::LspState::new(
+                    config.lsp_concurrency,
```
acheroncrypto: Please use the new [`concurrency_limit`](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/middlewares/limit.rs#L5) middleware instead:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/server/src/main.rs#L52

server/src/program.rs
```diff
@@ -29,6 +29,65 @@ pub const MAX_PATH_LEN: usize = 128;
...
+
+    if let Some(max_bytes) = max_bytes {
+        let bytes: usize = files.iter().map(|(_, content)| content.len()).sum();
+        if bytes > max_bytes {
+            return Err(anyhow!(
+                "Exceeded maximum project size: {bytes} > {max_bytes}"
+            ));
+        }
+    }
+
```
acheroncrypto: This code is not shared. It should only exist in the `lsp` code.

```suggestion
```

server/src/program.rs
```diff
@@ -29,6 +29,65 @@ pub const MAX_PATH_LEN: usize = 128;
...
+    Ok(())
+}
+
+fn is_valid_path(path: &str) -> bool {
+    path.len() <= MAX_PATH_LEN
+        && !path.contains("..")
+        && !path.contains("//")
+        && (is_cargo_file(path) || is_src_rust_file(path))
+}
```
acheroncrypto: Please inline this.

server/src/program.rs
```diff
@@ -29,6 +29,65 @@ pub const MAX_PATH_LEN: usize = 128;
...
+    {
+        Some(stem) => {
+            !stem.is_empty()
+                && stem
+                    .bytes()
+                    .all(|b| b.is_ascii_alphanumeric() || matches!(b, b'_' | b'-' | b'/'))
+        }
+        None => false,
+    }
+}
```
acheroncrypto: This doesn't seem more readable than the previous approach, but can keep if you don't want to have a single-line simple regex.

server/src/sandbox.rs
```diff
@@ -259,6 +259,8 @@ impl<'a> Sandbox<'a> {
                             .arg(cmd.get_program())
                             .args(cmd.get_args())
                             .env_clear()
+                            // Without `PATH`, `docker` resolves only via the OS fallback path, which lacks `/usr/local/bin` on macOS
+                            .envs(std::env::var_os("PATH").map(|path| ("PATH", path)))
```
acheroncrypto: This leaks host `PATH` environment variable to the sandboxed container, including when the OS is not macOS (e.g. in production).

Adding the env variable to `cmd` conditionally using `#[cfg(target_os = "macos")]` would be a lot less invasive.

server/src/setup.rs
```diff
@@ -8,6 +8,12 @@ const IMAGES_DIR: &str = "images";
 /// Setup the server.
 pub async fn setup() -> Result<()> {
     build_images().await?;
+
+    let killed = lsp::kill_leftovers().await?;
```
acheroncrypto: Should probably kill them on the first interrupt signal too. Otherwise they'll keep running indefinitely until the next restart.

server/src/setup.rs
```diff
@@ -8,6 +8,12 @@ const IMAGES_DIR: &str = "images";
...
+    if killed > 0 {
```
acheroncrypto: This being integer seems useless.

server/README.md
```diff
@@ -23,6 +23,24 @@ This will:
...
+| Variable | Default | Meaning |
+| --- | --- | --- |
+| `PG_LSP_CONCURRENCY` | `4` | Maximum concurrent sessions |
+| `PG_LSP_CPU_LIMIT` | `1` | CPU cores per session container |
+| `PG_LSP_MEMORY_LIMIT` | `4294967296` | Memory per session container, in bytes (4 GiB) |
+| `PG_LSP_IDLE_TIMEOUT` | `600` | Seconds without a client message before a session is closed |
+| `PG_LSP_MAX_LIFETIME` | `14400` | Seconds after which a session is closed regardless of activity |
+
+`PG_PAYLOAD_LIMIT` also caps the project size a session accepts, and only origins listed in `PG_CLIENT_URLS` may open one (browsers do not apply CORS to WebSockets). Containers carry the `solpg.lsp` label; leftovers from a crashed server are killed on the next start.
+
```
acheroncrypto: Please remove documenting `unstable` features in detail here before stabilization. If you want, include a single line about new features are being developed with `--features unstable`.

```suggestion
```

server/templates/anchor-1.1.2/programs/program/Cargo.toml
```diff
@@ -1,3 +1,6 @@
+# DO NOT EDIT: the dependency set is fixed by the build image. Any change
+# disables server builds and intellisense until it is reverted.
+
```
acheroncrypto: Please revert because `cargo` files will either be hidden in the client, or they'll be read-only.
```suggestion
```

acheroncrypto (changes requested):
Thanks a lot for doing this! This is a massive PR, so first review is server-only.

client/src/commands/build/build.ts
```diff
@@ -167,7 +167,15 @@ const getBuildFiles = () => {
     buildFiles.push([buildPath, content]);
   }
 
-  // TODO: Add `cargo` files
+  // Root `cargo` files decide which build template (toolchain image) the
+  // server uses; only the unstable route accepts them
```
acheroncrypto: Nit: Remove references to the server internals (e.g. build template, toolchain image) because the client does not need to know about the server internals; only accepting `cargo` files part is enough.

client/src/commands/build/build.ts
```diff
@@ -167,7 +167,15 @@ const getBuildFiles = () => {
...
+  if (PgSettings.experimental.unstable) {
+    const rootPath = PgExplorer.getProjectRootPath();
+    for (const name of ["Cargo.toml", "Cargo.lock"]) {
+      const content = files[PgCommon.joinPaths(rootPath, name)]?.content;
+      if (content !== undefined) buildFiles.push([name, content]);
+    }
```
acheroncrypto:
```suggestion
    for (const name of ["Cargo.toml", "Cargo.lock"]) {
      const content = PgExplorer.getFileContent(name);
      if (content !== undefined) buildFiles.push([name, content])
    }
```

client/src/components/Editor/Monaco/languages/rust/lsp/client.ts
```diff
@@ -0,0 +1,374 @@
...
+import type { Disposable } from "../../../../../../utils";
+
+/** Monaco language id for Rust */
+const LANGUAGE_ID = "rust";
+
+/** Owner of the markers this client sets */
+const MARKER_OWNER = "rust-analyzer-lsp";
+
+/** Route of the language server WebSocket on the playground server */
+const ROUTE = "/unstable/lsp";
```
acheroncrypto: `/unstable` part should not be hardcoded. It should change based on the `PgSettings.experimental.unstable` value, and ideally in the `PgServer`.

client/src/components/Editor/Monaco/languages/rust/lsp/client.ts
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * Requests the bridge answers itself; `sync` exists because `cargo check`
+ * reads the disk, not the LSP overlay.
+ */
+const BRIDGE = { open: "solpg/open", sync: "solpg/sync" };
+
+/** rust-analyzer configuration, sent at initialize and on request */
+const SERVER_CONFIG = {
+  // The container has no network; every dependency is cached at image build
+  cargo: { extraEnv: { CARGO_NET_OFFLINE: "true" } },
```
acheroncrypto: Why is this not set in the server? This is a recurring theme; client does not need to know about how server internals work. Everything except the public API is unknown to clients.

client/src/components/Editor/Monaco/languages/rust/lsp/protocol.ts
```diff
@@ -0,0 +1,221 @@
+/**
+ * The subset of the Language Server Protocol the editor uses.
+ *
+ * Kept local instead of depending on `vscode-languageserver-protocol`: the
+ * client only needs the wire shapes below, and the package pulls the whole
+ * protocol surface in for the sake of types.
```
acheroncrypto: Why not just include the types dep ([`vscode-languageserver-types`](https://www.npmjs.com/package/vscode-languageserver-types)) as a dev dependency?

I'm usually a big fan of inlining instead of depending on external packages, but in this case, the package has 0 dependencies and is maintained by Microsoft (which we already depend on).

client/src/components/Editor/Monaco/languages/rust/lsp/status.ts
```diff
@@ -0,0 +1,56 @@
...
+export const setStatus = (next: LspStatus) => {
+  if (next === status) return;
+  status = next;
+  PgCommon.createAndDispatchCustomEvent(ON_DID_CHANGE_STATUS, status);
+};
+
+/** Subscribe to status changes; the callback also runs immediately with the current value. */
+export const onDidChangeStatus = (
+  cb: (status: LspStatus) => unknown
+): Disposable =>
```
acheroncrypto: Nit: [`PgCommon.onDidChange`](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/client/src/utils/common.ts#L1060) always returns a `Disposable`.

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/CONTRIBUTING.md#L74

```suggestion
) =>
```

client/src/components/Editor/Monaco/languages/rust/lsp/status.ts
```diff
@@ -0,0 +1,56 @@
...
+/** Signal a message crossing the socket; throttled. */
+export const pulseActivity = () => {
+  const now = Date.now();
+  if (now - lastPulse < ACTIVITY_THROTTLE_MS) return;
+  lastPulse = now;
+  PgCommon.createAndDispatchCustomEvent(ON_DID_PULSE_ACTIVITY);
+};
+
+/** Subscribe to socket activity pulses. */
+export const onDidPulseActivity = (cb: () => unknown): Disposable =>
```
acheroncrypto: Same here:

```suggestion
export const onDidPulseActivity = (cb: () => unknown) =>
```

client/src/components/Editor/Monaco/languages/rust/lsp/workspace.ts
```diff
@@ -0,0 +1,91 @@
+import * as monaco from "monaco-editor";
+
+import { PgCommon, PgExplorer } from "../../../../../../utils";
+
+/** Files sent to the server: `[relative path, content]` pairs */
+export type Files = Array<[string, string]>;
```
acheroncrypto: We already have this type:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/client/src/utils/explorer/types.ts#L73-L74

You don't have to do it, but we should probably rename it to `Files` for consistency with the server. IIRC I didn't do it because of the potential ambiguity with [`ExplorerFiles`](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/client/src/utils/explorer/types.ts#L14). I will check this later.

client/src/components/Editor/Monaco/languages/rust/init.ts
```diff
@@ -1,3 +1,59 @@
...
+ * The server backend also restarts on workspace switch: one project per session.
+ */
+export const init = () => {
+  let active: Disposable | undefined;
+  let generation = 0;
+
+  const start = async (backend: typeof PgSettings.editor.rustAnalyzer) => {
+    const current = ++generation;
+    active?.dispose();
+    active = undefined;
```
acheroncrypto: Please use `null` here:

https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/CONTRIBUTING.md#L69

```suggestion
    active = null;
```

client/src/components/Editor/Monaco/languages/rust/init.ts
```diff
@@ -1,3 +1,59 @@
...
+        active = disposable;
+        if (backend === "server") setStatus("connected");
+      }
+    } catch (e) {
+      if (backend === "server" && current === generation) {
+        setStatus("disconnected");
+      }
+      const message = e instanceof Error ? e.message : String(e);
+      PgTerminal.println(
+        PgTerminal.error(`Rust Analyzer (${backend}): ${message}`)
```
acheroncrypto: Nit: We always include the word "error" if there was an error. For example, could prefix this text by `Process error:` for consistency with other processes. `PgTerminal.error` can be removed if you do this because it will get automatic highlighting.

Feel free to keep it as is if you think current one is better.

client/src/components/Editor/Monaco/languages/rust/init.ts
```diff
@@ -1,3 +1,59 @@
...
+  };
+
+  setRestartHandler(() => start(PgSettings.editor.rustAnalyzer));
+
+  // Setting change events fire once on subscription with the current value
+  PgSettings.onDidChangeEditorRustAnalyzer(start);
+  PgExplorer.onDidSwitchWorkspace(() => {
+    const backend = PgSettings.editor.rustAnalyzer;
+    if (backend === "server") start(backend);
+  });
```
acheroncrypto: `init` functions should not leak, meaning they should return a `Disposable` if they use change event listeners. Currently, there is an [issue](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/client/src/components/Editor/Monaco/Monaco.tsx#L684-L689) with the `Monaco` editor not disposing them, but that could change in the future.

Also, why not use [`PgCommon.batchChanges`](https://github.com/solana-playground/solana-playground/blob/f882778a00bc18aeaf62e8b3d2c3dfeb97d127ed/client/src/utils/common.ts#L1089)?

client/src/frameworks/anchor/files/Cargo.toml
acheroncrypto: Please remove this. I have a branch that handles this in another way.

client/src/frameworks/anchor/__tests__/cargo.test.ts
acheroncrypto: Same here; please remove.

client/src/frameworks/anchor/files/index.ts
```diff
@@ -1,6 +1,7 @@
 import type { TupleFiles } from "../../../utils";
 
 export const files: TupleFiles = [
+  ["Cargo.toml", require("./Cargo.toml")],
```
acheroncrypto:
```suggestion
```

client/src/settings/editor/editor.ts
```diff
@@ -7,4 +7,18 @@ export const editor = [
     default: "default" as const,
   }),
   createSetting({ id: "editor.wordWrap" }),
+  createSetting({
+    id: "editor.rustAnalyzer",
+    name: "Rust Analyzer",
```
acheroncrypto: Can omit:

```suggestion
```

client/src/settings/editor/editor.ts
```diff
@@ -7,4 +7,18 @@ export const editor = [
...
+    description:
+      "Where Rust intellisense runs. `Browser (WASM)` resolves against " +
+      "bundled Anchor 0.29 crates. `Server` uses the build server's " +
+      "toolchain, so it matches the project's template versions (requires " +
+      "a server started with `--features unstable`).",
```
acheroncrypto: Few potential improvements:

- Could mention WASM is offline
- Could mention that WASM uses the legacy template/crates instead of Anchor 0.29, as Anchor is not a requirement
- Could mention the `Server` option is the newer and better option instead of mentioning its internals (most ppl don't care how things work internally)
- Please either remove the server instructions or add a `TODO:` to remove them

client/src/setupTests.ts
acheroncrypto: Could you include this file in `__tests__` directory like you've been doing, and then point `jest` to it? Having an unconventional (naming-wise) test file at the root of the client is not very nice.

acheroncrypto (changes requested):
Client-side looks good overall. The actual impl looks solid. Most of my comments are nits.

